PR #4, opened 2015-11-16
allow runtime auth via socket.io

randomnerd:
you can see this functionality being used at my feathers-react-postgres demo here: https://github.com/randomnerd/feathers-react-pg

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
   "name": "feathers-authentication",
   "description": "Add Authentication to your FeathersJS app.",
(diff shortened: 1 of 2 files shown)


marshallswain:
Can you please set the `main` back to the `lib` folder?  The module uses ES6, now, so Babel builds the main folder upon install.

marshallswain:
I'm going to merge this and fix it.

randomnerd:
how do you use the package then? when i install it there is no lib folder so it doesnt work

PR #18, opened 2015-12-18
Fixing issue where next is called twice when hitting an unauthenticated service via socket.io

### LINKED ISSUE #17: next called twice using socket.io and using an unauthenticated service
It's called once from the checkToken function (it’s passed in as a the callback) then again at the bottom of the middleware since its not returning and continues on.

When I add the feathers-authentication middleware (and don’t have any hooks set up to force auth yet) then my FIND via sockets throws an error after receiving results. Commenting out the middleware and it works again.

corymsmith:
This fixes #17 and fixes up some intermittently failing tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
-  - 'node'
+  - '4.2.3'
   - 'iojs'
--- a/src/index.js
+++ b/src/index.js
@@ -134,21 +134,32 @@ export default function(config) {
       }
 
       // Socket.io middleware
-      if(io) {
+      if (io) {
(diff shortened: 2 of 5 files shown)


marshallswain:
It looks like the build is failing when trying to install the `bcrypt` module.  It looks like it's attempting on NodeJS version 5: https://travis-ci.org/feathersjs/feathers-authentication/jobs/97692661#L2823

Maybe that module hasn't been updated to work with Node v5.  Want to try setting up the travis.yml to use Node v4.2.2?

corymsmith:
Still fails on bcrypt, buffer-util and utf-8-validate modules. The same thing happened with my documentation change earlier this week.

PR #35, opened 2016-01-31
safety check for data

deanmcpherson:
If data isn't defined a TypeError is thrown (e.g. the username doesn't exist). Added a check so that the appropriate error is returned instead.


marshallswain:
Thanks for catching that!

PR #39, opened 2016-02-05
Validate if provider

mastertinner:
fix some hooks to not fire if the request is internal

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '4.2.3'
+  - '4.2.6'
   - 'iojs'
 before_install:
   - sudo apt-get install python-software-properties
--- a/src/hooks/hash-password.js
+++ b/src/hooks/hash-password.js
@@ -7,7 +7,8 @@ import bcrypt from 'bcrypt';
  */
 export default function(passwordField = 'password'){
   return function(hook) {
-    return new Promise(function(resolve, reject){
+
+    return new Promise(function(resolve, reject) {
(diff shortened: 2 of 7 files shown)


marshallswain:
For `queryWithUserId`, it would probably be better to remove the `else` and error throwing, which would remove the need for a provider check, I think.  That way, if a `hook.params.user` is in place, it will automatically add it.  Then it isn't duplicating functionality from the `requireAuth` hook, but works with it.  Here's an example that uses some of the new hooks in the upcoming OAuth release:

```
messageService.before({
  find: [
    authHooks.verifyToken({secret: 'feathers-rocks'}),
    authHooks.populateUser(),
    authHooks.requireAuth(),
    authHooks.queryWithUserId()
  ]
});
```

marshallswain:
I think the same could be said for `restrictToSelf` and `setUserId`.  Thanks for your help!

mastertinner:
thanks for your feedback, @marshallswain! i adapted the PR accordingly.

daffl:
@marshallswain Good to go? We can merge it. There may be some conflicts with the decoupling branch by @ekryski but I don't think the hooks changed too much.

@mastertinner Thank you for contributing!

ekryski:
:+1: Ya we can merge this stuff. I did make a couple changes to hooks but I'll sort out the conflicts.

Mad props to @mastertinner for his help! :boom:

marshallswain:
Thanks for taking care of this, y'all.  I'm a little spread thin right now.  :)
Thanks @mastertinner for your help.

mastertinner:
thanks for the support, guys!

Issue #44, opened 2016-02-07
Add a client side component for authentication


ekryski:
Since Feathers is isomorphic we should follow a similar convention to some of our other libs and expose a client side portion.

Auth can get pretty complicated. We don't need to go so far as implementing everything [Auth0's Lock](https://auth0.com/lock) provides but it would be nice to have something that at a minimum provides:

- a mechanism for managing storing a valid token (ie. localStorage)
- handling a 401 error
- refreshing tokens
- refreshing and storing the logged-in user's data
- exposes a client side hook for adding the token to the header or socket body
- handles pulling the JWT from the `feathers-jwt` cookie after successful OAuth authentication

### Things that need to happen to work on this

- [x] Add header support for feathers REST client
- [x] Fix providers to work with new params format
    - [x] REST
    - [x] socket.io
    - [x] primus
- [x] Add ability to determine which providers are enabled in the client
    - [x] REST
    - [x] socket.io
    - [x] primus
- [x] Pass `params.query` instead of `params` into `makeUrl` ([see this](https://github.com/feathersjs/feathers-rest/blob/master/src/client/base.js#L12))
- [x] Add a default service hook if feathers authentication is enabled that attempts to grab the JWT from local storage (falling back to cookie) and stores it on `hook.params.token`
- [x] allow configuring some custom options for client side `feathers-authentication` module
    - a custom user endpoint  (default is `/users`)
    - a custom auth header field  (default is `Authorization`)
    - custom token endpoint (default is `/auth/token`)
    - custom local auth endpoint (default is `/auth/local`)

ekryski:
Once #48 is merged it's as simple as:

```js
var socket = io(host, {
  transport: ['websockets']
});

window.app = feathers()
  .configure(feathers.socketio(socket))
  .configure(feathers.hooks())
  .configure(feathers.authentication());

app.io.on('connect', function(){
  app.authenticate({
    type: 'local',
    'email': 'admin@feathersjs.com',
    'password': 'admin'
  }).then(function(result){
    console.log('Authenticated!', result);
  }).catch(function(error){
    console.error('Error authenticating!', error);
  });
});
```

PR #48, opened 2016-02-11
Adding an auth client

ekryski:
This adds an authentication client that makes it simple to authenticate with feathers.

Using it is as simple as:

```js
var socket = io(host, {
  transport: ['websockets']
});

window.app = feathers()
  .configure(feathers.socketio(socket))
  .configure(feathers.hooks())
  .configure(feathers.authentication());

app.io.on('connect', function(){
  app.authenticate({
    type: 'local',
    'email': 'admin@feathersjs.com',
    'password': 'admin'
  }).then(function(result){
    console.log('Authenticated!', result);
  }).catch(function(error){
    console.error('Error authenticating!', error);
  });
});
```

### included

- `app.authenticate()` - authenticates a user over rest or sockets using local or token auth. This automatically checks localstorage with fallback to cookie for the JWT stores the token and user in local storage upon success.
- `app.logout()` - logs out a user destroying the user and token in local storage
- `app.user()` - a convenience method for getting the current user
- token is included in `hook.params.token` so it can be access by hooks and services on the client and sent to the server.
- user is included in `hook.params.user` so it can be access by hooks and services on the client and sent to the server.


ekryski:
This refs #44

src/client/index.js
```diff
@@ -0,0 +1,188 @@
...
+      if (options.type === 'local') {
+        endPoint = authOptions.localEndpoint;
+      } else if (options.type === 'token') {
+        endPoint = authOptions.tokenEndpoint;
+      }
+      else {
+        throw new Error(`Unsupported authentication 'type': ${options.type}`);
+      }
+
+      // return new Promise(function(resolve, reject) {
```
ekryski: I left this in here to see if @daffl feels like taking a crack at Promise hell. The intention is to see if we can make it so the developer doesn't have to wrap `app.authenticate()` in a callback after socket connection.

marshallswain:
Well, that's one hell of a pretty API. :+1:   Nice work, @ekryski.

Issue #25, opened 2016-01-18
Handling expired tokens


marshallswain:
In the [middleware that checks the Authorization header](https://github.com/feathersjs/feathers-authentication/blob/master/src/index.js#L100), if the token is expired we are currently sending back a `401 Unauthorized` response.  I don't think that should be handled here.  If it's an unprotected resource that doesn't require auth, shouldn't we treat it the same as if it were some anonymous user?  I think that a `requireAuth` hook should probably handle whether or not that request is rejected.  Any thoughts otherwise?

marshallswain:
I'm going to fix this so that all authorization is handled through hooks.  When an expired token is provided it will allow the request to go through as it would for an anonymous user instead of punishing those who have expired tokens.

ekryski:
At first I wasn't sure but I think I know what you are getting at here. In the `decoupling` branch I've moved this to a hook called `verifyToken` that can be registered as opposed to blanket middleware. It is currently registered as a before hook on the `/auth/token` service.

My question though now that I think about is, we probably want to always check to see if there is a token (like you were doing) and attempt to validate it only if it is there. Otherwise just move on to the next hook.

The reason being is that the end user would use `req.user` in middleware or `params.user` in hooks to do any authorization but because we aren't using sessions you need to "authenticate" the user on every request.

So my proposal is actually that we have middleware that fires before any other, that attempts to authenticate the user if a token was passed in, otherwise it just moves on through.

marshallswain:
That makes sense to me.  We can use a `decodeToken` hook that does only that, then use the `requireAuth`- or `requireAuthForPrivate`-type hooks to control access.

ekryski:
Ya that what I was thinking but `decodeToken` would be middleware that is applied to **any** route. Basically global middleware, otherwise we'd have to explicitly add a `decodeToken` hook to each service the developer wants to protect.

Now that I type that out, maybe that's not such a big deal???

marshallswain:
Yeah.  It's not a big deal to do
```
service.before({
  all: [
    hooks.auth.decodeToken(), 
    hooks.auth.requireAuth()
  ]
});
```
I think having it be explicit, and not automatic, will be more helpful in the long run.

ekryski:
k :+1:. That's makes it super easy then. If that's the case then we could just use:

```js
service.before({
  all: [authHooks.verifyToken()]
```

You wouldn't even need the `requireAuth` call.

marshallswain:
[...]

Issue #45, opened 2016-02-08
Support graceful fallback to cookies


ekryski:
Currently when using OAuth, in order to get the JWT we generate back to the client, we shove it in a cookie called `feathers-jwt`. The client then parses the JWT from the cookie and uses this JWT to authenticate with the API going forward.

In theory, in addition to checking the `Authorization` header, query sting, or request/socket `body` we could also check the `feathers-jwt` cookie to see if it is expired and if not grab the token.

Now that I think about this, this is actually pretty easy and should just be a couple lines added [right in this middleware](https://github.com/feathersjs/feathers-authentication/blob/decoupling/src/middleware/index.js#L52).

ekryski:
This is actually done in the `decoupling` branch now. It doesn't support cookies on sockets however...

Issue #43, opened 2016-02-07
Support OAuth2


ekryski:
We should leverage passport rather than write our own OAuth negotiation code. When a user authenticates we return the user id and a Feathers JWT so that the JWT can be used for further API access.

There are quite a few social sign on providers out there that support OAuth2. Most notably, Facebook, Github, Google, etc.

We can leverage the `passport-<provider>` modules and let the user pass in the `Strategy` and simply map the strategy to an OAuth2 service. This will get us very far, very quickly.

ekryski:
This is working in the [decoupling](https://github.com/feathersjs/feathers-authentication/tree/decoupling) branch and is slated to merged pretty soon.

Issue #38, opened 2016-02-04
Only sign the JWT with user id. Not the whole user object


ekryski:
If we sign with the whole user object the JWT can get really big.

marshallswain:
Need to update the docs once this is in place.
http://docs.feathersjs.com/authorization/readme.html#getting-your-user-schema-just-right

ekryski:
Will do. I've fixed this in the `decoupling` branch.

PR #49, opened 2016-02-12
Decoupling

ekryski:
This brings in a pretty major overhaul to feathers authentication. With it comes:

- Migrating existing code to use services
- Standardizing on a hook spec:

  ```js
  export default function(options = {}){
    const defaults = {passwordField: 'password'};
    options = Object.assign({}, defaults, options);

    return function(hook) {
      // do stuff
    };
  }
  ```

- Adds support for authenticating with socketio and primus (#32)
- Only signs the JWT with user id (#38)
- Locks down socket authentication (#33)
- Continues the work @marshallswain did on handling expired tokens (#25)
- Adds a bunch more tests.
- Adds support for OAuth2 (#43)
- Adds a client side component for easy authentication with Feathers (#44)
- Adds preliminary support for graceful fallback to cookies for JWT (#45)
- Adds an example project showing all the different ways you can authenticate


src/client/index.js
```diff
@@ -0,0 +1,188 @@
...
+      if (options.type === 'local') {
+        endPoint = authOptions.localEndpoint;
+      } else if (options.type === 'token') {
+        endPoint = authOptions.tokenEndpoint;
+      }
+      else {
+        throw new Error(`Unsupported authentication 'type': ${options.type}`);
+      }
+
+      // return new Promise(function(resolve, reject) {
```
daffl: Does this commented out code need to kept for reference?

ekryski: @daffl No we can get rid of it unless you want to take a crack at it.

src/client/index.js
```diff
@@ -0,0 +1,188 @@
...
+    app.logout = function() {
+      // remove user and token from localstorage
+      // on React native it's async storage
+      utils.clearToken();
+      utils.clearUser();
+    };
+
+    // Set up hook that adds adds token to data sent to server over sockets
+    app.mixins.push(function(service) {
+      service.before(hooks.populateParams());
```
daffl: It looks like this needs hooks configured in the app for the authentication plugin to work at all. Should we throw an error if `service.before` and `service.after` isn't a function?

src/hooks/verify-token.js
```diff
@@ -0,0 +1,34 @@
...
+ * @param {String} options.secret - The JWT secret
+ */
+export default function(options = {}){
+  const secret = options.secret;
+
+  return function(hook) {
+    const token = hook.params.token;
+
+    if (!token) {
+      return Promise.resolve(hook);
```
daffl: Technically you could just `return hook;`

src/services/local/index.js
```diff
@@ -0,0 +1,121 @@
...
+        // Handle bad username.
+        if (!user) {
+          return done(null, false);
+        }
+
+        return user;
+      })
+      .then(user => {
+        // Check password
+        bcrypt.compare(password, user[this.options.passwordField], function(error, result) {
```
daffl: I'm wondering if we should move promisify from https://github.com/feathersjs/feathers-socket-commons/blob/master/src/utils.js#L17 into feathers-commons and use this here. It would turn this section from 12 LOC into 2.

src/client/index.js
ekryski: Ya I just had that thought a couple minutes ago.

src/services/oauth2/index.js
```diff
@@ -0,0 +1,186 @@
...
+
+        let data = Object.assign({
+          [`${options.provider}Id`]: profile.id,
+          [`${options.provider}`]: profile._json
+        });
+        
+        return app.service(options.userEndpoint).create(data, { internal: true }).then(user => {
+          return done(null, user);
+        }).catch(done);
+      }).catch(done);
```
daffl: From 60 to here it can probably just be changed to 

```js
return app.service(options.userEndpoint).create(data, { internal: true })).then(user => done(user)).catch(done);
```

src/services/oauth2/index.js
```diff
@@ -0,0 +1,186 @@
...
+    let app = this.app;
+    
+    // TODO (EK): Make this configurable
+    if (id !== 'callback') {
+      return Promise.reject(new errors.NotFound());
+    }
+
+    return new Promise(function(resolve, reject){
+    
+      let middleware = passport.authenticate(options.provider, authOptions, function(error, user) {
```
daffl: This is where `promisify` would come in handy, too. Not super urgent but might make things easier to follow and less callback-y

src/services/local/index.js
ekryski: Yup. We could do that. I'm going to iterate on this repo a bit more over the next few days so I don't think we should delay on merging but if we get that util in commons then I'll use it here. It would actually come in handy in a few spots.

src/hooks/verify-token.js
ekryski: true. I still have tests to write for these hooks so I'll do that at the same time.

src/client/index.js
daffl: I actually didn't read all of it ;) What would it have to do?

src/services/oauth2/index.js
ekryski: I don't think so. Passport expects the first param to `done` to be an error or null.

src/services/local/index.js
daffl: I was going to add to the comment that it's not super urgent but would be a nice refactoring just to make things a little more terse.

src/services/oauth2/index.js
ekryski: Ya totally.

daffl:
Looks like it needs to be merge with master which is hopefully not too much of a disaster. Besides that, let's merge it, make a 0.2.0 release and iterate from there.

src/client/index.js
ekryski: That's the promisfying we were trying to do the other night so that a developer doesn't have to call `app.io.on('connect', cb)` before they can authenticate. We would handle that for them. I couldn't get it to work. Too many promises and event listeners.

ekryski:
@daffl I had already did that and resolved conflicts. It's good there.

All the tests pass locally all the time. I dunno why that one test keeps timing out randomly. I can try setting the mocha timeout higher to see if that works.

ekryski:
never mind. The latest build did pass on all three platforms now that I added the timeout in the test. I think for some reason the tests were running fast enough that the `expiredToken` wasn't actually expired yet.

ekryski:
This PR. Closes #25, #32, #33, #38, #43, #44, #45.

Issue #64, opened 2016-02-15
Find "query" is replaced by token


fastlorenzo:
Hello, 
In : feathers-authentication/src/client/hooks.js 
`export let populateSocketParams = function() {
  return function(hook) {
    if (hook.params.token) {
     ** hook.params.query = {
        token: hook.params.token
      };**
    }
  };
};`

You should create a new object and keep the existing params, otherwise every param query passed to the 'find' action is replaced by the token only.

Thanks :+1:

daffl:
@ekryski Is there a reason for adding the token to the query? Shouldn't it be already set on the socket?

fastlorenzo:
A way to fix it could be using the object.assign call

Lorenzo Bernardi

marshallswain:
I think this might be related to this issue: https://github.com/feathersjs/feathers/issues/224

marshallswain:
Oh.  it's on the query.  nm.

fastlorenzo:
Changing 
`hook.params.query = { token: hook.params.token };`
to
`hook.params.query = Object.assign({}, hook.params.query, {
        token: hook.params.token
      });`
fixes the issue.
Could you integrate the fix or should I create a pull request ?

marshallswain:
Would you mind creating a PR?  Thanks.

ekryski:
@fastlorenzo thanks for the PR. Let me review. I think it's likely that it's my bad but I want to make sure it doesn't break anything. I haven't had time to write all the tests that I've needed to.

fastlorenzo:
@ekryski sure, no issue :+1:

fastlorenzo:
PS: maybe you should also use the Object.assing for the following line in the same file : 
`19: hook.params.headers = {`

ekryski:
@fastlorenzo ya I think you are right on the headers part. I actually think @daffl might be right and that for sockets we don't need to set the token each time. `hook.params.query` and `hooks.params.data` are the only things sent to the server by feathers-client but in the case of sockets because it is a persistent connection we should already have the token on server side.

fastlorenzo:
Indeed, and I think sending the token each time in the query can be an issue, for example if you want to filter the query using a parameter named `token`.

fastlorenzo:
@ekryski I quickly checked by removing the token to be sent in each request and it look ks like it's working, as @daffl said.
I'll create another pull request for that so you can check it.

fastlorenzo:
[...]

PR #65, opened 2016-02-15
Fixing client query

fastlorenzo:
Fixing issue #64

### DIFF
--- a/src/client/hooks.js
+++ b/src/client/hooks.js
@@ -11,14 +11,14 @@ export let populateHeader = function(options = {}) {
   const defaults = {
     header: 'Authorization'
   };
-  
+
   options = Object.assign({}, defaults, options);
 
   return function(hook) {
     if (hook.params.token) {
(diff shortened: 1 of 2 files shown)


ekryski:
It looks good. I've got some changes coming in a PR that does the same thing plus some others on the back end side that you are probably unaware of and also adds some tests. Just give me another 15 minutes or so.

ekryski:
Thanks so much for your help @fastlorenzo! I'm going to close this in favour of #67 because there is actually a bit more than needed to happen to clean up dead code and I thought it would be faster if I just do it rather than explaining everything that needed to be removed and why.

ekryski:
Fuck it. I'll merge it to give credit where it is due. Thanks @fastlorenzo!

marshallswain:
Atta boy!  :)

fastlorenzo:
Haha thanks @ekryski

PR #67, opened 2016-02-15
Removing assigning token to params.query for sockets.

ekryski:
Much thanks to @fastlorenzo for pointing this issue out and providing a PR (#65).

There is actually a bit more code to clean up than was done in his PR so I've continued his work with this one. I also wanted to add a few more automated tests to make sure things are still working.


ekryski:
Closes #65.

PR #68, opened 2016-02-16
Add module exports Babel module and test CommonJS compatibility

daffl:
This pull request adds the Babel export module and verifies CommonJS compatibility of the built module.


ekryski:
Looks all good. :shi

PR #76, opened 2016-02-24
Refactoring for storage service

ekryski:
Initial refactoring and tests for client side authentication that works with a storage service. In order to support caching of tokens and the current user on React Native, NodeJS and the browser we needed an abstraction over top of localStorage, AsyncStorage. So it made sense to create a service adapter that abstracts the underlying storage mechanism. We've done this with https://github.com/feathersjs/feathers-localstorage.

This will be able to be merged once https://github.com/feathersjs/feathers-localstorage/pull/1 lands and then we will be able to use `feathers-authentication` in React Native and Node. :smile:


ekryski:
Still a WIP. Opened for visibility. Do not merge yet.

ekryski:
Pending tests passing this should be good to merge. Has been manually tested with React Native and the browser.

PR #77, opened 2016-02-24
Adding support for OAuth2 token based auth strategies. Closes #46.

### LINKED ISSUE #46: Support mobile authentication via OAuth2
The OAuth stuff that is about to land will work just fine on mobile if you are inside a browser. However, if you are talking to feathers from a native app or from React Native you won't be able to follow the redirects that are required for the OAuth flow. Therefore, we need to optionally support token based social auth like [`passport-facebook-token`](https://github.com/drudge/passport-facebook-token) where you would already have the access token from authenticating with the OAuth provider on the mobile device.

My thought is that a developer can include the this strategy and pass it as a `tokenStrategy` option. If that option is passed we set up that token strategy (potentially along side the regular strategy) and then map that to the `create` service method (ie. `POST /auth/facebook`).

ekryski:
This adds support for authenticating via OAuth tokens if a `tokenService` is configured by `POST`ing an `auth_token` to a Feathers OAuth service. For example:

```
POST /auth/facebook { "auth_token": "your token" }
```

This allows a mobile (or some other device) to not have to embed the `clientID` and `clientSecret` and just let the server manage that. Most of the time this is when a mobile client does authentication between apps.

An example of a token strategy would be: https://github.com/drudge/passport-facebook-token.

### DIFF
--- a/src/services/oauth2/index.js
+++ b/src/services/oauth2/index.js
@@ -144,9 +144,7 @@ export class Service {
                   .catch(reject);
       });
 
-      middleware(params.req, params.res, function() {
-        console.log('I was called', arguments);
-      });
+      middleware(params.req, params.res);
     });
   }
 
--- a/src/services/oauth2/index.js
+++ b/src/services/oauth2/index.js
@@ -29,8 +29,6 @@ export class Service {
       }
     };
 
-    // console.log('Authenticating', accessToken, refreshToken, profile);
-
     // Find or create the user since they could have signed up via facebook.
     app.service(options.userEndpoint)
       .find(params)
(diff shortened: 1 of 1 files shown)


src/services/oauth2/index.js
```diff
@@ -108,39 +108,47 @@ export class Service {
...
+        };
+
+        // Get a new JWT and the associated user from the Auth token service and send it back to the client.
+        return app.service(options.tokenEndpoint)
+                  .create(payload, { internal: true })
+                  .then(resolve)
+                  .catch(reject);
+      });
+
+      middleware(params.req, params.res, function() {
```
daffl: This was probably just for debug right?

ekryski: lol yup

ekryski:
Going to roll this in as a minor release because the generator, generates code that needs this in order to work. If there is any client side specific stuff that needs to happen for React Native we can deal with that in another PR.

PR #79, opened 2016-02-24
bumping feathers-errors version

ekryski:


ekryski:
tests passing locally. Very benign change.

PR #87, opened 2016-02-29
update populateUser() hook

kulakowka:
- get auth options from app config;
- use idField from  auth options for get user by id

I'm not sure that everything is done correctly. It would be great if the author of the hook will make review for this PR.


ekryski:
@kulakowka nice! This was on my list of todos, so thanks! Looks good.

I'm finally sitting down to write test for this stuff so I'll merge it in and fix anything if it's not working as expected.

Issue #7, opened 2015-11-16
Password-less Local Auth with Email / SMS


marshallswain:
Be able to send an email or text with an auth link.

ekryski:
I'm working on this right now actually and it's coming along nicely.

daniel79:
Would be good to have option to implement own authentication provider (for example instead of using 'local' we clould use 'myServiceName')

pierretasci:
I kind of mimicked something like this in my own app where I create a wrapper service that handles auth requests in place of the token service provided by this repo. If the user is new or logging in, it redirects to an email login service (using passwordless) that handles the email and redirect workflow.

When the user clicks the email link, it attempts to authenticate again using the token in the email. This request is again sent to the wrapper but this time it forwards it to the token service. It isn't super pretty but it gets this done while still using JWT tokens as the primary logged in auth scheme.

Hope that helps.

ekryski:
Wow time flies. I've been busy fixing bugs and writing tests. I have a stashed branch with passwordless auth working. Might need a couple tweaks due to the updates that just landed in auth but I'm hoping to have something up soon.

daffl:
Related to https://github.com/feathersjs/feathers-authentication/pull/96

/cc @ccummings

philippnagel:
@ekryski Any progress on your implementation? Just asking before implementing everything myself.

ekryski:
See https://github.com/feathersjs/feathers-authentication-management. I haven't tried it yet but will next week. I believe it is very close to release. Feel free to give it a go and report any issues on that repo.

cc/  @eddyystop

I'm going to close this as that is the going to be the repo for account confirmation, password reset, etc. going forward.

rhythnic:
I created an example repo for doing passwordless auth with `feathers-authentication-management`.

https://github.com/rhythnic/feathers-passwordless-auth-example

PR #96, opened 2016-03-08
Adding passwordless service

ccummings:
Adds passwordless auth via the [passwordless](https://passwordless.net/) library.

This is setup to work stateless without sessions to match the rest of feather's auth, but that can be enabled with the `sessionSupport` middleware if needed in the future.

Authentication flow works like this:

- User submits email address to `auth/passwordless`
    - If the user is new, create user
    - Generate a token and deliver it using the various delivery methods (email, sms, carrier pigeon)
- User clicks the link they were given which should `GET auth/passwordless`
    - We validate the token and return a JWT token (using the token service)

At this point, all auth is done with the JWT token.

I've added a `passwordless` service and made a few adjustments to the middleware. I also created an example which works for authenticating.

It seems the feathers client stuff around authentication and authorization is not working right now (retrieving messages throws errors), but the service itself is modeled closely after `local` and `token` so when that is sorted, this should work just fine.


ekryski:
Holy :hankey: @ccummings! This looks great! I'll take a look at this today and hopefully get it merged in pretty quickly.

ekryski:
This would close #7 and probably #5.

daffl:
There are some timeouts in the tests. Not sure why those are happening.

ccummings:
After playing with this a bit more, I'm not sure the `passwordless` library is the best fit. It does some odd things like closing the response so you get "headers already sent" errors.

Something like this could be made more composable so it could be used in a number of different schemes. to do that it should:
- Generate different kinds of tokens (algorithmic, simple pin codes, etc)
- Store them using existing feathers DB adapters
- Deliver them via email, sms, etc using feathers service plugins
- verify tokens and redirect to other services

This would enable the normal passwordless flow (link in email), multi factor auth (pin code to SMS or code to email after a local auth) or could be used for generating tokens for forgotten passwords.

Still trying to sort out exactly how this would look, open to suggestions.

ekryski:
@ccummings that was my thought when skimming it that it was doing a lot more than we need. You nailed exactly what I've been thinking. I think it's actually pretty easy. I'm going to be digging into auth again today so I will push a PR tonight/tomorrow for this functionality.

jonjamz:
Excited for this feature!

daffl:
Did we reach a verdict on this?

ekryski:
I think we've agreed that using passwordless isn't the right solution because it does a lot more than we need. That being said I'd like to leave it open for reference until I get my PR up. Hopefully in the next couple days busy with some other stuff at the moment.

ccummings:
This is much better off being written from the ground up using services and middleware instead of trying to shoehorn an existing middleware solution which doesn't seem to play nice with other middleware.

@ekryski I'd be happy to help out on this if you have a branch in the works.

daffl:
Sounds good. Glad everybody agrees. I'm going to close this PR then. Thanks again for adding this @ccummings, hopefully we can iterate to the final version.

PR #97, opened 2016-03-09
fix html in templates

mastertinner:


ekryski:
Sweet! Thanks @mastertinner!

PR #99, opened 2016-03-09
Ten hours is only 36 seconds

mileswilson:
The cookies expiration date is currently set for 36 seconds in the future, not ten hours.


ekryski:
Uhhhhh yup. Thanks @mileswilson!

PR #101, opened 2016-03-13
Finalizing client side authentication module

### LINKED ISSUE #95: Client should store token string and not the token object
I thought I had fixed this but it looks like you need to reference `hook.params.token.value` in https://github.com/feathersjs/feathers-authentication/blob/master/src/client/hooks.js#L22.

daffl:
This is a refactoring for the client side authentication to finalize the API and make it easier to use:

```js
const app = feathers()
  .configure(socketio(socket))
  .configure(hooks())
  .configure(authentication())

// Store the token in local- or async storage
const app = feathers()
  .configure(socketio(socket))
  .configure(hooks())
  .configure(authentication({
    storage: window.localStorage // AsyncStorage
  }));

// Authenticate with local username and password
app.authenticate({
  type: 'local',
  email: 'hello@feathersjs.com',
  password: 'supersecret'
}).then(response => {
  console.log('Authenticated', response);

  app.get('token') // -> the JWT
  app.get('user') // -> the user
});

// Authenticate with cookie JWT or JWT from storage
app.authenticate().then(response => {
  console.log('Authenticated', response);

  app.get('token') // -> the JWT
  app.get('user') // -> the user
});
```

Closes #95


daffl:
This should be done for now. Tested for REST, Socket.io and Primus. One tests is skipped because I couldn't find out how to disconnect an existing Primus socket for the life of me.

src/client/utils.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+}
+
+// Returns a storage implementation
+export function getStorage(storage) {
+  if(storage) {
+    return storage;
+  }
+
+  return {
```
ekryski: What's this part for? Just a shim for localstorage so that it will work with node without having to include another module?

test/client/index.test.js
```diff
@@ -1,73 +1,220 @@
 import assert from 'assert';
```
ekryski: :+1: :+1: on this whole file. Nice tests.

src/client/utils.js
daffl: Yep. Just an in-memory shim for the localStorage API. I used to have `window.localStorage` in there as well but I think it's better to do it explicitly.

test/client/index.test.js
```diff
@@ -1,73 +1,220 @@
...
+        }).then(() => {
+          return app.authenticate().then(response => {
+            assert.deepEqual(app.get('token'), response.token);
+            assert.deepEqual(app.get('user'), response.data);
+          });
+        }).then(done);
       }).catch(done);
+  });
+
+  it.skip('.logout works, lets you not access to protected service', done => {
```
ekryski: Was this one meant to be skipped? It's not working?

src/client/index.js
```diff
@@ -1,137 +1,99 @@
...
+      service.before(hooks.populateParams(config));
     });
-    
+
     // Set up hook that adds authorization header for REST provider
     if (app.rest) {
       app.mixins.push(function(service) {
-        if (!service.before || !service.after) {
-          throw new Error(`It looks like feathers-hooks isn't configured. It is required before you configure feathers-authentication.`);
+        if (typeof service.before !== 'function' || typeof service.after !== 'function') {
```
ekryski: I realized that we probably don't need this second duplicate warning because it should be caught by the check above.

test/client/index.test.js
ekryski: Ah I saw the note above. Ok you disconnect a socket by calling `.end()` I believe.

daffl: Ah. Yeah, the problem was that this test fails with sockets because `app.logout` really has to disconnect the socket, too. I'll give it a shot with `.end()`.

src/client/index.js
daffl: True that. I'll remove it.

ekryski:
:shipit:

PR #104, opened 2016-03-15
retrieve user id when populateUser() is an after hook

### LINKED ISSUE #103: `idField` breaks from `tokenService.create()` to `populateUser()` after hook
hey,

so [ `tokenService.create()` returns `{ id: user[idField] }`](https://github.com/feathersjs/feathers-authentication/blob/master/src/services/token/index.js#L96-L98), but [`populateUser()` after hook expects `{ [idField]: id }`](https://github.com/feathersjs/feathers-authentication/blob/master/src/hooks/populate-user.js#L24-L27). this means `populateUser()` isn't doing anything after `tokenService.create()`, which means [the same for `localService.create()`](https://github.com/feathersjs/feathers-authentication/blob/master/src/services/local/index.js#L78-L79).

i'm not sure what's the expected behavior here so i haven't made a pull request, but this lack of any user data returning from `client.authenticate()` breaks an upgrade from `0.2`, or maybe i'm missing something entirely.

ahdinosaur:
here's a minimal change to fix #103.

### DIFF
--- a/src/hooks/populate-user.js
+++ b/src/hooks/populate-user.js
@@ -23,7 +23,7 @@ export default function(options = {}){
 
     // If it's an after hook grab the id from the result
     if (hook.type === 'after') {
-      id = hook.result[options.idField];
+      id = hook.result[options.idField] || hook.result.id;
     }
     // Check to see if we have an id from a decoded JWT
     else if (hook.params.payload) {


ekryski:
The issue is actually this one: https://github.com/feathersjs/feathers-authentication/issues/107.

The problem is [this line](https://github.com/feathersjs/feathers-authentication/blob/master/src/services/token/index.js#L97) should be `[options.idField]: user[options.idField]`.

So this isn't really the right fix. Took me a while to figure out WTF was causing the problem. Your PR actually pointed me in the right direction but I'm to close this one. Thanks Mikey!

ahdinosaur:
ah yep @ekryski, i thought of changing that line as a potential solution but wasn't sure if it was intentional to always be `id`. in hindsight i reckon you're right, cheers. :+1:

ekryski:
Ya it looks like it just got missed when we started landing PRs to propagate auth config options down to the services and hooks.

PR #106, opened 2016-03-16
primus client connect event is 'open'

ahdinosaur:
see https://github.com/primus/primus#events

### DIFF
--- a/src/client/utils.js
+++ b/src/client/utils.js
@@ -18,7 +18,8 @@ export function connected(app) {
 
     // If the socket is not connected yet we have to wait for the `connect` event
     if( (app.io && !socket.connected) || (app.primus && socket.readyState !== 3)) {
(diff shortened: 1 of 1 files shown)


daffl:
Thank you! I was wondering about that actually. Do you have an idea why the Primus tests was passing before?

ekryski:
@ahdinosaur good catch. I totally missed that when I review the PR. That's weird that those tests were passing. We'll need to look at that. This is good to merge though.

ahdinosaur:
@daffl is it because [the tests only run after the 'open' event](https://github.com/feathersjs/feathers-authentication/blob/master/test/integration/primus.test.js#L46-L51)?

daffl:
Oh. Haha, well that'll do it. I thought it was timing out but it was totally a workaround for that bug. I'm pretty sure we can take that out now.

PR #115, opened 2016-03-23
client should return a 401 error code when no token is provided

ccummings:
If no token is provided to the `authenticate` call, a 401 code should be returned with the error.

This makes the [jQuery](https://github.com/feathersjs/feathers-docs/blame/master/guides/jquery.md#L196) and [React](https://github.com/feathersjs/feathers-docs/blame/master/guides/react.md#L246) guides redirect when a user lands on the chat.html without logging in.


ekryski:
Ah shit. I already fixed this in my branch. I'll merge this and resolve conflicts. Thanks @ccummings!

Issue #138, opened 2016-03-29
Make all hooks optional if used internally


ekryski:
Most of the hooks already do this. These ones need to be updated:

- [x] query with current user
- [x] associate with current user
- [x] hash password

daffl:
It is possible for a user to overwrite `params.provider` so you run the risk of exposing unsecured methods to the outside world. Although unlikely we might want to investigate using `Object.defineProperty` so that it doesn't happen accidentally.

ekryski:
Sure. Let's create another issue for that though because that touches all the hooks and I don't want to block shipping in this in order to prevent someone from _possibly_ screwing up their hooks.

Issue #136, opened 2016-03-28
logout should de-authenticate a socket


ekryski:
Currently the token is stored on the socket when you successfully authenticate. Calling logout currently doesn't remove this token from the socket on the server side. It only removes it from localStorage on the client side. Since the token is not sent with every request using sockets, logging out doesn't take effect unless the socket is disconnected by either the client or the server.

Issue #132, opened 2016-03-26
[Security] JsonWebToken Lifecycle Concerns; Set HttpOnly = true in JWT cookie


petermikitsh:
Currently the JsonWebToken ("JWT") is returned [without the HttpOnly flag](https://github.com/feathersjs/feathers-authentication/blob/fca25f2b5b03b773975595ae9ac034199df5127b/src/middleware/index.js#L96). Consequently, client side script can access the JWT. This introduces a theoretical XSS vulnerability. 

Consider the security implication of allowing client side access to the JWT in this scenario:

----

A bad actor somehow injects a script a client-side script into your app (how it got in could be a wide range of possibilities, given the flexible nature of feathers framework. e.g., somehow some bad client JS code ends up being inserted in a database, and is now served in all client requests). In the script, the hacker inspects localStorage, read the `feathers-jwt` cookie (if httpOnly is false), and send a POST request to www.badactor.com/xss, and now have a copy of the end user's JWT. The hacker can then issue requests to the application server, authenticated as the end user.

----

This example illustrates the broader concern of authentication architecture within Feathers, specifically the lifecycle of the JSON Web Token. It's critical to be highly specific in the implementation to mitigate certain attack vectors. JWT's are permitted to be transported and stored through a number of options in feathers, e.g., headers, cookies, query parameters, localStorage, etc. A unique set of security concerns are introduced to each application. The [documentation](http://docs.feathersjs.com/authentication/readme.html#authentication-over-rest) does address the tradeoffs of the transports.

If you set HttpOnly to true, and enforce an architectural pattern of limiting the JWT to cookies, you can effectively mitigate XSS. If the JWT is placed in localStorage, the XSS vector isn't mitigated because localStorage can be read by client side script (which is kind of the point of localStorage, making it a less than ideal place for JWT's to live).

marshallswain:
Doesn't putting the token in a cookie actually make JWT less secure?  Cookies are automatically sent with every request, so now we're back to being vulnerable to the attacker not needing to even get access to the JWT token in order to make API calls.

daffl:
The cookie value will only be set once when loading the app after a login and I'm not sure it can be HTTP only because the JS app needs to access it in order to make requests using the JWT. We decided that sending it in the cookie is still safer than the current general practice of putting it in the callback query string.  I am also not sure if there is any place other than LocalStorage where the JWT could be stored. 

[...]

Issue #128, opened 2016-03-26
restrictToRoles hook needs to throw an error and not scope the query


ekryski:
Same issue as #127 but different hook.

PR #139, opened 2016-03-30
0.7 Release

ekryski:
This PR has a bunch more bug fixes and some pretty critical security fixes

- [x] Lock down cookie (#132)
- [x] Add unit tests for cookie options
- [x] can now use default redirect routes with a custom handler (#121)
- [x] Add middleware tests for successfulLogin
- [x] Add middleware tests for failedLogin
- [x] Prevent emitting auth service events (#126)
- [x] Add tests to make sure auth service events are not fired
- [x] RestrictToOwner hook needs to throw an error (#128)
- [x] RestrictToRoles hook needs to throw an error (#127)
- [x] OAuth user profile should be updated (#124)
- [x] All hooks should support internal usage passthrough (#138)
- [x] Clear cookie on logout (#122)
- [ ] Update docs, examples and generator
- [x] de-auth socket on logout (#136)
- [x] get logout integration tests working
- [x] Move to bcryptjs instead of native brcrypt
- [x] Removes ability to authenticate with the cookie that is used to transmit the JWT to the client


src/client/index.js
```diff
@@ -1,8 +1,16 @@
...
+  connected,
+  authenticateSocket,
+  logoutSocket,
+  getJWT,
+  getStorage,
+  clearCookie
+} from './utils';
 
 const defaults = {
+  cookie: 'fathers-jwt',
```
daffl: It's like your dad's JWT :wink:

marshallswain: Yep.  JWT on punch cards.

src/client/utils.js
```diff
@@ -36,9 +36,27 @@ export function authenticateSocket(options, socket, method) {
...
+// Returns a promise that de-authenticates a socket
+export function logoutSocket(socket, method) {
+  return new Promise((resolve, reject) => {
+    // If we don't get a logged out message within 10 seconds
+    // consider it a failure. 
+    const timeout = setTimeout(function() {
+      reject(new Error('Could not logout over socket'));
+    }, 10000);
+
+    socket.once('logged out', function() {
```
daffl: Couldn't this just be the `logout` event?

src/client/utils.js
```diff
@@ -50,13 +68,22 @@ export function getCookie(name) {
   return null;
 }
 
+// Returns the value for a cookie
+export function clearCookie(name) {
+  if (typeof document !== 'undefined') {
+    document.cookie = `${name}=;expires=Thu, 01 Jan 1970 00:00:01 GMT;`;
```
daffl: Is this a standard default?

src/hooks/restrict-to-owner.js
```diff
@@ -30,13 +34,17 @@ export default function(options = {}){
...
-    // current user, which will result in a 404 if they are not the owner.
-    hook.params.query[options.ownerField] = id;
-    
-    // TODO (EK): Maybe look up the actual document in this hook and throw a Forbidden error
-    // if (field && id && field.toString() !== id.toString()) {
-    //   throw new errors.Forbidden('You do not have valid permissions to access this.');
-    // }
+    // look up the document and throw a Forbidden error if the user is not an owner
+    return new Promise((resolve, reject) => {
+      this.get(hook.id).then(data => {
```
daffl: Do we need to add `params` here? `this.get(hook.id, hook.params)`?

src/index.js
```diff
@@ -21,13 +21,21 @@ const PROVIDERS = {
...
+  shouldSetupFailureRoute: true,
   successRedirect: '/auth/success',
   failureRedirect: '/auth/failure',
   tokenEndpoint: '/auth/token',
   localEndpoint: '/auth/local',
   userEndpoint: '/users',
   header: 'authorization',
-  cookie: 'feathers-jwt'
+  cookie: {
+    enabled: true,
```
daffl: Can this just be considered disabled when someone set `cookie: false`?

src/index.js
```diff
@@ -145,6 +143,24 @@ export default function auth(config = {}) {
     // Register error handling middleware for redirecting to support
     // redirecting on authentication failure.
     app.use(middleware.failedLogin(authOptions));
+
+    // Setup route handler for default success redirect
+    if (authOptions.shouldSetupSuccessRoute) {
```
daffl: What was the reason again for adding those? I mean if you do a redirect it will almost certainly be to your own success page no?

src/middleware/index.js
```diff
@@ -183,6 +186,20 @@ export let setupSocketIOAuthentication = function(app, options = {}) {
         }).catch(errorHandler);
       }
     });
+
+    socket.on('logout', function() {
```
daffl: Ah ok I see why there is two events. But you can also send a callback from the client `socket.on('logout', callback => callback())` so you don't need two different events (that's what we're doing for calling service methods.

src/services/local/index.js
```diff
@@ -111,6 +112,11 @@ export default function(options){
     // Get our initialize service to that we can bind hooks
     const localService = app.service(options.localEndpoint);
 
+    // prevent regular service events from being dispatched
+    if (app.io || app.primus) {
```
daffl: I'd probably go with `if(typeof localService.filter === 'function')`. We should probably also add warnings if someone tries to override those methods.

src/services/oauth2/index.js
```diff
@@ -179,6 +182,11 @@ export default function(options){
     // Get our initialized service
     const service = app.service(options.endPoint);
 
+    // prevent regular service events from being dispatched
```
daffl: Actually this can probably even go in the services `setup`

test/client/index.test.js
```diff
@@ -119,10 +119,17 @@ const setupTests = initApp => {
         expect(response.data).to.not.equal(undefined);
 
         app.logout().then(() => {
+          expect(app.get('token')).to.equal(null);
+          expect(app.get('user')).to.equal(null);
+
           app.service('messages').create({ text: 'auth test message' })
             .then(msg => console.log('!!!', msg))
```
daffl: `console.log`

daffl: That looks like I left it in there ;)

test/integration/rest.test.js
```diff
@@ -329,14 +329,14 @@ describe('REST authentication', function() {
...
-          request(options, function(err, response) {
+          request(options, function(error, response) {
             expect(response.statusCode).to.equal(401);
             done();
           });
         });
 
-        it('returns error instead of data', (done) => {
-          request(options, function(err, response, body) {
+        it('returns erroror instead of data', (done) => {
```
daffl: Global replace?

src/client/index.js
ekryski: LOL. Man I always make that typo. Cracks me up every time.

src/client/utils.js
ekryski: It just needs to be set in the past, so it can be whataver.

src/client/utils.js
ekryski: no, but we could use a callback instead of two events.

src/hooks/restrict-to-owner.js
ekryski: Probably ya. Good catch!

src/index.js
ekryski: I think we can probably do that.

src/index.js
ekryski: The problem is if someone wants to use the same route name but a different handler. Which, I think is actually probably the most common use case. You can't rely on just inspecting whether the `successRedirect` is different than the default.

src/middleware/index.js
ekryski: Ha! I had started doing it that way but then I switched to be consistent with the `authenticate` and `authenticated` events. I think callbacks is probably a better idea.

test/client/index.test.js
ekryski: Ya. It never gets there now but I'll remove it.

test/integration/rest.test.js
ekryski: yes thank you!

src/client/utils.js
daffl: :+1:

src/index.js
daffl: What I'm saying is why are there default auth redirect HTML pages at all? I might be missing something but they don't seem particularly useful. If you do a browser redirect it'll pretty much always be to your app right? Couldn't we make it required and throw an error if it isn't set (or disabled)?

src/services/local/index.js
ekryski: What were you thinking of for a warning?

```js
if (typeof tokenService.filter === 'function') {
      tokenService.filter(() => false);
    } else if (app.io || app.primus) {
      throw new Error('You cannot override the filter method');
    }
```

How can we detect whether the filter event behaves the way it should other than testing to make sure that events don't get dispatched?

daffl: Nothing in here, it'll be added for socket-commons and feathers-hooks.

src/services/oauth2/index.js
ekryski: I had to think about this for a second but I think you are right. It **should** go in there because then someone can override setup if they really need that functionality for whatever reason. We might come back to this when we address scaling services.

src/index.js
ekryski: For web apps you are correct, you would pretty much always override them. However, if you use the pop-up method or are on mobile, with the defaults you don't need to do anything to the server. You check that you navigated to `/auth/success` grab the JWT from the cookie and close the pop-up window or the webview on mobile.

src/services/local/index.js
ekryski: :+1:

ekryski:
Ok this is good to go! Just need to merge some docs PRs and update the generator to use the `restrictToAuthenticated` hook.

ekryski:
This isn't actually broken it just depends on https://github.com/feathersjs/feathers-hooks/pull/64.

PR #140, opened 2016-03-30
Fix copy paste typo in queryWithCurrentUser hook.

juodumas:


juodumas:
Btw., restrictToOwner and queryWithCurrentUser hooks are identical.

ekryski:
Thanks @juodumas! I pushed 0.7.0 today so now `restrictToOwner` is quite different.

PR #167, opened 2016-04-20
[WIP] Add Google and Facebook authentication endpoints

beevelop:
This is a PoC of enabling "stateless" client-side authentication for Google and Facebook. It simply exposes the `google` and `facebook` authentication types and maps it to the appropriate endpoints.

Use case for this manual and decoupled authentication workflow are primarily mobile applications (e.g. Cordova-based solutions like Ionic, etc.) and single-page applications in general. Additionally I personally like the resulting (more consistent) authentication API and a "Socket.io-only" solution (as there are no need for additional REST calls):
```js
feathers.authenticate({
  type: 'google',
  access_token: 'no42..aBcdef-gH42_...iJklM_4N_Op'
}).then(function (result) {
  console.log('Authenticated!', result)
  console.log(feathers.get('user'))
}).catch(function (error) {
  console.error(error)
})
```

## Edit: Google-Sidenote
To get google-token working I additionally had to fork `passport-google-token` (https://github.com/beevelop/passport-google-token). The current implementation fails quite silently if there is no refreshToken passed in the body (as it checks for the `req.query` and `req.headers` afterwards – what results in an unhandled error)

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -13,7 +13,9 @@ const defaults = {
   cookie: 'feathers-jwt',
   tokenKey: 'feathers-jwt',
   localEndpoint: '/auth/local',
-  tokenEndpoint: '/auth/token'
+  tokenEndpoint: '/auth/token',
+  facebookEndpoint: '/auth/facebook',
+  googleEndpoint: '/auth/google'
 };
 
 export default function(opts = {}) {
@@ -56,6 +58,10 @@ export default function(opts = {}) {
           endPoint = config.localEndpoint;
         } else if (options.type === 'token') {
           endPoint = config.tokenEndpoint;
+        } else if (options.type === 'facebook') {
+          endPoint = config.facebookEndpoint;
+        } else if (options.type === 'google') {
+          endPoint = config.googleEndpoint;
         } else {
           throw new Error(`Unsupported authentication 'type': ${options.type}`);
         }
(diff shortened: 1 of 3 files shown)


ekryski:
@beevelop Nice. I think what we'd probably want to do is make it less hardcoded and behave a bit more like how the backend side sets up the routes and stuff. Maybe calling it like this instead?

```js
app.authenticate({
  type: 'oauth2',
  provider: 'google', // endpoint ends up /auth/google
  accessToken: 'xxxxx' 
}).then(function(result) {
  // do your stuff
}).catch(function(error) {
  // handle your error
});
```

this can then dynamically create the `/auth/<provider>` endpoint. Additionally I think we should allow someone to pass a custom endpoint that would override the dynamically generated one. Like so:

```
app.authenticate({
  type: 'oauth2',
  provider: 'google',
  accessToken: 'xxxxx' ,
  endPoint: '/api/auth/google' // overrides the default /auth/google endpoint
}).then(function(result) {
  // do your stuff
}).catch(function(error) {
  // handle your error
});
```js

ekryski:
Related to #169

beevelop:
I definitely agree that the whole workflow could be a lot more decoupled. Do you think this is in the scope of this PR?
Considering your suggestion: In what way would `type: 'oauth2'` be used in the authentication flow?

kristianmandrup:
Please fix issues and merge :)

beevelop:
@ekryski: Any update on this?

ekryski:
@beevelop sorry for dropping the ball on this. This stuff makes sense on the client side for sure. I'm going to take some of the same ideas and cherry pick them into the 0.8.0 branch we've got going.

ekryski:
I'm going to close this PR but if you want @beevelop we can keep discussing #226 about how this might work. I think I'll actually be able to get this support in later this week.

Issue #175, opened 2016-04-25
restrictToOwner should not throw an error on mass deletions


thosakwe:
Hi guys, really enjoy the work you guys have done. Just one problem.

See, I made this package, [feathers-seeder](https://github.com/thosakwe/feathers-seeder). A key part of its functionality is being able to delete all records from a service if the configuration object dictates to do so.

However, this is prevented by this section of code in `src/hooks/restrict-to-owner.js`:
https://github.com/feathersjs/feathers-authentication/blob/master/src/hooks/restrict-to-owner.js#L15-17

```js
if (!hook.id) {
      throw new errors.MethodNotAllowed(`The 'restrictToOwner' hook should only be used on the 'get', 'update', 'patch' and 'remove' service methods.`);
}
```

So if I call the following, the operation will fail:

```js
service.remove(null, additionalParams);
```

As you can imagine, this renders feathers-seeder more or useless. 

I am going to submit a PR that fixes this. Instead of checking if the hook has an `id`, the code should be changed to see if the `method` matches 'get', 'update', 'patch' or 'remove'.

thosakwe:
Here is the [PR](https://github.com/feathersjs/feathers-authentication/pull/176).

ekryski:
Providing the ability to conditionally run hooks (#210) should solve this problem.

ekryski:
We actually came up with a better solution. You can simply wrap any existing hooks in your own, like so:

```js
import { hooks } from 'feathers-authentication';

exports.hashPassword = function(options) {
  // Add any custom options

  function(hook) {
    return new Promise((resolve, reject) => {
      if (myCondition !== true) {
        return resolve(hook);
      }

      // call the original hook
      hooks.hashPassword(options)(hook)
        .then(hook => {
          // do something custom
          resolve(hook);
        })
        .catch(error => {
          // do any custom error handling
          error.message = 'my custom message';
          reject(error);
        });
    });
  });
}
```

This provides the ultimate flexibility and doesn't require a change to the core hooks.

PR #176, opened 2016-04-25
Fixed bug in restrictToOwner

thosakwe:
See issue [#175](https://github.com/feathersjs/feathers-authentication/issues/175).

### DIFF
--- a/src/hooks/restrict-to-owner.js
+++ b/src/hooks/restrict-to-owner.js
@@ -12,7 +12,7 @@ export default function(options = {}){
       throw new Error(`The 'restrictToOwner' hook should only be used as a 'before' hook.`);
     }
 
-    const methodIsAllowed = hook.method === 'get' || hook.method === 'update' || hook.method === 'patch' || hook.method == 'remove';
+    const methodIsAllowed = hook.method === 'get' || hook.method === 'update' || hook.method === 'patch' || hook.method === 'remove';
     if (!methodIsAllowed) {
       throw new errors.MethodNotAllowed(`The 'restrictToOwner' hook should only be used on the 'get', 'update', 'patch' and 'remove' service methods.`);
     }
--- a/src/hooks/restrict-to-owner.js
+++ b/src/hooks/restrict-to-owner.js
@@ -12,7 +12,8 @@ export default function(options = {}){
       throw new Error(`The 'restrictToOwner' hook should only be used as a 'before' hook.`);
     }
 
-    if (!hook.id) {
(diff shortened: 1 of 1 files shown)


daffl:
This looks like a good fix but `npm test` is failing for that hook now.

ekryski:
Thanks for the PR @thosakwe! Unfortunately I don't think it's the right fix. Currently the hook assumes `hook.id` is present. Only changing the error check will just cause the hook to [fail later on](https://github.com/feathersjs/feathers-authentication/blob/master/src/hooks/restrict-to-owner.js#L44) by returning a 404.

I actually think it is very valid that is throws an error on mass deletion. If you want to support mass deletion then I think there are 3 different options:

- you should use a custom hook
- use the `restrictToRoles` hook and set a user role that has access to do mass deletions
- we should support an optional callback to determine when the hook should execute, like we do in the other bundled hooks (see [here](http://docs.feathersjs.com/hooks/bundled.html#disable))

In order to determine whether you have access to items you want to modify you need to look them up first, this could be incredibly expensive if you did a blank find on all items for the service. So in reality, I think what you are looking for is to conditionally disable or enable this hook based on some condition.

thosakwe:
Ah, I see.

And on second though, my fix might have been a bit ill-advised, too. Not sure if it would be a good to tell [feathers-seeder](https://github.com/thosakwe/feathers-seeder) users, "Hey! To use this plug-in, you have to remove this security feature from your application."

ekryski:
@thosakwe well, like I said you can just change your hooks or put in the readme for feathers-seeder that you do need to change the hook.

It sounds like the more ideal solution is to allow conditional running of the hooks.

ekryski:
I'm going to close this as I think we should support conditionally running hooks. I've created #210 to track that.

shadowRR:
I agree about the fix, but this hook isn't returning the proper error. 

It should handle both case, when you are on the wrong hook method, and when the id is missing. Otherwise, you can loose some time trying to figure out why feathers tells you you're using the hook on the wrong method, while it isn't the case.

UlyssesAlves:
I don't know if it's related, but I'm receiving an error saying **The 'restrictToOwner' hook should only be used on the 'get', 'update', 'patch' and 'remove' service methods.**. This error ocurs when I try to search for a user by id inside a middleware funcion:

`app.get('/notificacoes/email/alteracao-de-senha/:idUsuario/:idSolicitacao', notificacoes.processeSolicitacaoDeAlteracaoDeSenha);`

servicoDeUsuarios.get(idUsuario)
            .then(usuario => {
                  // code here is not run
             });
             catch(error => {
                  // code enters inside this error handler with the above said message.
             });

What is wrong with this code, please?

PR #183, opened 2016-04-27
fix(oauth2): Use patch to update user in oauthCallback

### LINKED ISSUE #174: Facebook Authentication should do a patch not an update.
Using a feathers-cli generated app with mongodb, if the user model has more fields (other than facebook-id and facebook data), all those fields gets erased when the user authenticates. 

This happens because the authentication proccess updates the social information of the user by sending a full update to the user service, thus eliminating any other field the user may have. I think should be better to patch the user instead of updating it.

beevelop:
Using update() leads to conflicts with non-default user models and effectively in loosing the user's
data. By using patch the data gets merged properly.

fixes #174

### DIFF
--- a/src/services/oauth2/index.js
+++ b/src/services/oauth2/index.js
@@ -60,7 +60,7 @@ export class Service {
 
           debug(`Updating user: ${id}`);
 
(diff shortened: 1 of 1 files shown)


ekryski:
Thanks @beevelop!

Issue #171, opened 2016-04-21
Optionally Include password in the params.query object passed to User.find()


toddgeist:
I have a DB that requires userName and Password to verify the user record.  The user service only receives the userName now.  The password is left off.  I understand that it isn't necessary for most systems. But it is for mine. 

How about a config option to include password in params.query?

If I am not mistaken, a little code could be added right here
https://github.com/feathersjs/feathers-authentication/blob/master/src/services/local/index.js#L22-L26

maybe something like this?
```
const params = {
   query: {
        [this.options.usernameField]: username
    }
};
//optionally add the password to params.query
if(this.options.includePasswordInQuery){
   params.query[this.options.passwordField] = password
}
```
Would you consider a pull request with this?

Thanks

daffl:
I think there already was a feature request somewhere to be able to query for different fields. But there still needs to be a password comparison function. How are you comparing your passwords? I assume they aren't just plain text.

toddgeist:
In this DB The passwords are stored hashed, but I can't access the hash.  All I can do is query the DB with a UserName and Password, if the Password matches I get back the User Record.  If it doesn't I do not.

So I query the DB, if the UserRecord comes back _I hash the password right then_ and stick it on the userObject.  Then the normal comparison function is happy and generates the token.

A bit of hack I'll grant you but what else can I do :-)

toddgeist:
#165 seems related

Issue #165, opened 2016-04-18
Pass more to local authentication params


saiichihashimoto:
I'm trying to create an app where there's different areas of the room, somewhat like slack teams. A person may have the same email in different areas. So when someone logs in, I want to retrieve them by username but ALSO by the area. In other words, I need to add to the [params used to retrieve from the user service](https://github.com/feathersjs/feathers-authentication/blob/master/src/services/local/index.js#L22). As it stands, there's no way to do that.

ekryski:
I think the only way would be to add a **before** hook on your user service that adds the params when going through authentication. The problem with the `checkCredentials` function signature is that is dictated by passport, I don't think that we can monkey with it.

Did you have any success @saiichihashimoto?

saiichihashimoto:
A before hook doesn't solve my problem. When check credentials looks up the user via the app service, it only passes the username as a query parameter. I want to use information from the request (in my case, the current subdomain), and have that be one of the query parameters. Regardless of any before hooks we'll always get the same query, currently.

Just like #171, I need to do some logic on the parameters before I pass them in. I'm fine hacking together a solution for my specific case, but I'm having a hard time doing what I want without having to rewrite the whole file for my case.

saiichihashimoto:
Does what I'm saying make sense?

PR #186, opened 2016-04-28
Allow manipulation of params before checking credentials

saiichihashimoto:
This allows manipulating the parameters of the user service before checking the credentials.

Closes feathersjs/feathers-authentication#165 Closes feathersjs/feathers-authentication#171

### DIFF
--- a/src/services/local/index.js
+++ b/src/services/local/index.js
@@ -18,12 +18,10 @@ export class Service {
     this.options = options;
   }
 
-  checkCredentials(username, password, done) {
-    const params = {
-      query: {
-        [this.options.usernameField]: username
-      }
-    };
+  checkCredentials(req, username, password, done) {
+    let params = this.options.checkCredentialsParams ? this.options.checkCredentialsParams(req, username, password) : {};
+    params.query = params.query || {};
+    params.query[this.options.usernameField] = username;
 
     // Look up the user
     this.app.service(this.options.userEndpoint)
@@ -105,7 +103,7 @@ export class Service {
 }
 
 export default function(options){
-  options = Object.assign({}, defaults, options);
+  options = Object.assign({ passReqToCallback: true }, defaults, options);
   debug('configuring local authentication service with options', options);
 
   return function() {
--- a/src/services/local/index.js
+++ b/src/services/local/index.js
@@ -10,6 +10,7 @@ const debug = Debug('feathers-authentication:local');
 const defaults = {
   usernameField: 'email',
   passwordField: 'password',
+  passReqToCallback: true,
   session: false
 };
 
@@ -113,7 +114,7 @@ export class Service {
 }
 
 export default function(options){
-  options = Object.assign({ passReqToCallback: true }, defaults, options);
+  options = Object.assign({}, defaults, options);
   debug('configuring local authentication service with options', options);
 
   return function() {
--- a/src/services/local/index.js
+++ b/src/services/local/index.js
@@ -18,14 +18,24 @@ export class Service {
     this.options = options;
   }
 
-  checkCredentials(req, username, password, done) {
-    let params = this.options.checkCredentialsParams ? this.options.checkCredentialsParams(req, username, password) : {};
-    params.query = params.query || {};
-    params.query[this.options.usernameField] = username;
+  buildCredentials(req, username) {
+    const usernameField = this.options.usernameField;
+    return new Promise(function(resolve) {
+      const params = {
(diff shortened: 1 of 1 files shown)


marshallswain:
@saiichihashimoto thanks.  Can you give an example of how to use this?

saiichihashimoto:
Sure! In my case, I allows users to create different "organizations", which they interact with in a different subdomain. There is only one user db across all the organizations, however. So, when setting up `feathers-authentication`, you'd do something like:

```javascript
app.configure(authentication({
  local: {
    checkCredentialsParams: function(req, username, password) {
      var params = { query: {} };
      // Add the organization onto params.query
      return params;
    }
  }
));
```

`feathers-authentication` would go on to add the user onto the query and do its checking. In the case of #171, since @toddgeist wants to add the password onto the request, he'd now be able to do that.

src/services/local/index.js
```diff
@@ -105,7 +103,7 @@ export class Service {
 }
 
 export default function(options){
-  options = Object.assign({}, defaults, options);
+  options = Object.assign({ passReqToCallback: true }, defaults, options);
```
marshallswain: What does `passReqToCallback` do?

marshallswain:
I think this is a step in the right direction.

src/services/local/index.js
saiichihashimoto: If you look to `checkCredentials`, the first parameter is now the request.

Passport mentions it in http://passportjs.org/docs/configure#association-in-verify-callback

> One downside to the approach described above is that it requires two instances of the same strategy and supporting routes.
> 
> To avoid this, set the strategy's passReqToCallback option to true. With this option enabled, req will be passed as the first argument to the verify callback.

saiichihashimoto:
I'd imagine there's a cleaner (and more in-line with the rest of feathers) way to accomplish this, but its somewhat of a blocker for me and I'd love to have this feature. ;-)

src/services/local/index.js
ekryski: Ya this is pretty common when dealing with passport, especially if you need to do account consolidation. We're already doing this in the OAuth2 service. @saiichihashimoto let's move this up into the `defaults` object, that way all the defaults stay organized together.

ekryski:
@saiichihashimoto Yes this definitely is a step in the right direction so thanks for putting this together!

I think the best way for Feathers to stay flexible and accommodate all the different uses cases people have is to simply document more fully how you would extend an auth service. Other people have reported issues where they would like to do other things.

In reality someone should only need to extend the auth service and implement their own `checkCredentials` method. Then you can do whatever you need to inside of it. This is effectively the same as if you had to implement the Strategy callback with Passport. So what I suggest is that:

- we norm on a consistent method name across the auth services, maybe `authCallback`
- we allow you to extend the service
- we keep the `passRequestToCallback: true` in the default options

Make sense?

marshallswain:
Maybe we can add another promise attribute that resolves with the params that you want, and you can just replace the promise with whatever you want.  


```
buildCredentials(req){
  return new Promise(function(resolve, reject){
    let params = {
      username: req.body.username,
      password: req.body.password
    };
    resolve(params);
  }
}

checkCredentials(req, done) {
  this.buildCredentials(req).then(params => {
    // Look up the user
    this.app.service(this.options.userEndpoint).find(params)
    ...
  });
}
```

Then you can just provide your own buildCredentials.

We can of course make it more pure, this is just to illustrate the idea.  :)

ekryski:
@marshallswain I do like that!

marshallswain:
Cool!  (that was fast)  @saiichihashimoto do you want to modify your PR?  If not I will probably get to this on Monday or Tuesday.

saiichihashimoto:
This should cover it! Let me know if its what you're looking for.

marshallswain:
Looks good to me.  @ekryski you call the ship.

ekryski:
👍  thanks @saiichihashimoto! We'll need to document this but I have [an open issue](https://github.com/feathersjs/feathers-docs/issues/159) in the docs already. 🍻

ekryski:
I'll push a release over the weekend. It's been a while since I've looked at auth and I think some PRs have gone in to master already that aren't released so I need to see what changed.

saiichihashimoto:
Let me know when this happens @ekryski, really looking forward to it!

ekryski:
@saiichihashimoto published v0.7.6! Thanks again for the PR!

PR #197, opened 2016-05-03
properly handle optional `_json` property

### LINKED ISSUE #196: OAuth2 authentication callback failing due to missing property
I'm getting and error `TypeError: Cannot set property 'accessToken' of undefined` in https://github.com/feathersjs/feathers-authentication/blob/master/src/services/oauth2/index.js#L44 probably due to `_json` property is missing. Looking at other Strategies' implementation I see this in `passport-facebook` (set here: https://github.com/jaredhanson/passport-facebook/blob/master/lib/strategy.js#L179), but not, for example, in `passport-google`. My Strategy provider does not set it as well.

Can this be made conditional?

nyaaao:
fixes #196

### DIFF
--- a/src/services/oauth2/index.js
+++ b/src/services/oauth2/index.js
@@ -41,11 +41,11 @@ export class Service {
         // TODO (EK): This is where we should look at req.user and see if we
         // can consolidate profiles. We might want to give the developer a hook
         // so that they can control the consolidation strategy.
-        profile._json.accessToken = accessToken;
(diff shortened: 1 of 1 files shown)


nyaaao:
@daffl @ekryski does this look good to you guys?

ekryski:
@nyaaao yup this is great! Thanks! 😄

ekryski:
v0.7.7 published with the fix. Thanks again @nyaaao!

PR #208, opened 2016-05-18
Use removeItem in the storage on logout

### LINKED ISSUE #203: Remove token from localstorage
I'm wondering if the token should be removed and not just set to empty string ?

daffl:
`removeItem` is supported by both, LocalStorage and AsyncStorage so this PR adds it to the dummy storage and uses it when logging out (instead of setting it to an empty string).

Closes #203

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -80,9 +80,9 @@ export default function(opts = {}) {
       app.set('token', null);
 
       clearCookie(config.cookie);
-      
+
       // remove the token from localStorage
-      return Promise.resolve(app.get('storage').setItem(config.tokenKey, '')).then(() => {
+      return Promise.resolve(app.get('storage').removeItem(config.tokenKey)).then(() => {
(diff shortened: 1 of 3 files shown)


ekryski:
I'll need to double check RN but I'm pretty sure this is a much better fix than the empty string hack we had to add because AsyncStorage was barfing on `null` values.

daffl:
@ekryski Got a chance to try it? AsyncStorage [does support removeItem](https://facebook.github.io/react-native/docs/asyncstorage.html#removeitem) so I don't think it would cause any issues with RN.

ekryski:
I didn't get a chance to check it. Running into issues getting the chat app to build but I'm pretty sure this is fine. :shipit: 

We can revert if it's no good. I'll check it in a couple days.

daffl:
This has been released with v0.7.8. I think it should work but let's monitor to see it we get any error report regarding this.

PR #209, opened 2016-05-18
Small refactoring to simplify structure and remove code duplication

daffl:
I went through the code to refresh my memory and made some small tweaks for organization (move lonely `index.js` files in the `services` folder one level up), split the middleware file into one for Express and one for websockets and refactored the code duplication out of the websocket authentication handlers.

I don't think anything controversial but @ekryski should definitely double check.


ekryski:
This is great! Stuff I had been wanting to do but hadn't had time. Thanks @daffl!

PR #214, opened 2016-05-30
jwt ssl warning

aboutlo:
- Fire a warning if the request is insecure rather than refusing to send the cookie. 
- Allow override a single cookie property 

Please have a look https://github.com/feathersjs/feathers-authentication/compare/master...aboutlo:ls-JWT-ssl-warning?diff=unified&expand=1&name=ls-JWT-ssl-warning#diff-1fdf421c05c1140f6d71444ea2b27638L60

In my opinion `app.get('auth')` is always undefined at this stage.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,7 +32,7 @@ const defaults = {
   cookie: {
     name: 'feathers-jwt',
     httpOnly: false,
-    secure: process.env.NODE_ENV === 'production' ? true : false
+    secure: process.env.NODE_ENV === 'production'
   }
 };
 
@@ -56,9 +56,13 @@ export default function auth(config = {}) {
       config.local = {};
     }
 
+    if (config.cookie ){
+      config.cookie = Object.assign({}, defaults.cookie, config.cookie);
+    }
+
     // Merge and flatten options
-    const authOptions = Object.assign({}, defaults, app.get('auth'), config);
-    
+    const authOptions = Object.assign({}, defaults, app.get('auth'),config);
+
     // If a custom success redirect is passed in or it is disabled then we
     // won't setup the default route handler.
     if (authOptions.successRedirect !== defaults.successRedirect) {
@@ -83,7 +87,7 @@ export default function auth(config = {}) {
       // Get the token and expose it to REST services.
       app.use( middleware.normalizeAuthToken(authOptions) );
     }
-    
+
     app.use(passport.initialize());
 
     app.setup = function() {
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -32,7 +32,7 @@ const defaults = {
   cookie: {
     name: 'feathers-jwt',
     httpOnly: false,
-    secure: process.env.NODE_ENV === 'production' ? true : false
+    secure: process.env.NODE_ENV === 'production'
```
ekryski: 👍 Thank you!

src/index.js
```diff
@@ -56,8 +56,12 @@ export default function auth(config = {}) {
       config.local = {};
     }
 
+    if (config.cookie ){
+      config.cookie = Object.assign({}, defaults.cookie, config.cookie);
+    }
+    
     // Merge and flatten options
-    const authOptions = Object.assign({}, defaults, app.get('auth'), config);
```
ekryski: I think we do need this back. The reason being is you can just do `app.configure(authentication())` if you want to use all the default options **plus** your auth config defined in `default.json` or `production.json`. For example, your token secret, etc.

src/middleware/express.js
```diff
@@ -81,26 +81,28 @@ export function successfulLogin(options = {}) {
       // clear any previous JWT cookie
       res.clearCookie(options.cookie.name);
 
-      // Only send back cookies when not in production or when in production and using HTTPS
-      if (!req.secure && process.env.NODE_ENV === 'production') {
-        console.error(`You should be using HTTPS in production! Refusing to send JWT in a cookie`);
+      // Check HTTPS and cookie status in production 
+      if (!req.secure && process.env.NODE_ENV === 'production' && options.cookie.secure) {
+        console.warn('WARN: Request isn\'t served through HTTPS: JWT in the cookie is exposed.');
```
ekryski: Suweeet!

ekryski:
@aboutlo this is awesome! Nice work! I think we do need `app.get('auth')` from what I explained above. I probably need to add a test for that. If you can add that back in this is good to go! 🍻

aboutlo:
Thank you @ekryski for the kind words. I added back `app.get('auth')`. Tests are still green :)

daffl:
Oh sorry, this was totally good to go to. I'll merge it in, it'll make it into the next release.

ekryski:
Yup :shipit:

aboutlo:
cool

ekryski:
Released v0.7.9 with this fix in! 🍻 @aboutlo!

PR #219, opened 2016-06-03
Lowercase custom header

### LINKED ISSUE #218: Server-side header option does not accept capital letters
#### Issue

RESTful authentication fails if the optional `header` key in the [server-side configuration options](http://docs.feathersjs.com/authentication/readme.html#options) includes capital letters.

#### Background

I brought this up in the [Slack channel](https://feathersjs.slack.com/archives/general/p1464967390004911) yesterday and @ekryski asked that I open an issue with steps to reproduce the problem. This happens because internally Node's http module lower cases all headers names to normalize their values. If a user attempts to configure their authentication options with a header that includes capital letters lookup inside [`src/middleware/express.js`](https://github.com/feathersjs/feathers-authentication/blob/master/src/middleware/express.js#L35) fails.

For example, this configuration:

```js
// Application setup
.configure(authentication({ header: 'X-Auth-Token' })
```

Causes the following situation internally wen a POST request is made with a `X-Auth-Token` header:

```js
// Express middleware
console.log(options.header);              // => 'X-Auth-Token'
console.log(req.headers['x-auth-token']); // => token
console.log(req.headers['X-Auth-Token']); // => undefined
```

#### Steps to reproduce

1. Clone the repository locally and `cd` into the project root.

1. Open `example/app.js` in your text editor.

1. Add the optional `header` value with inside `example/app.js`:

  ```diff
diff --git a/example/app.js b/example/app.js
index 444ad12..9c52a6b 100644
--- a/example/app.js
+++ b/example/app.js
@@ -17,7 +17,8 @@ var app = feathers()
   .use(bodyParser.urlencoded({ extended: true }))
   // Configure feathers-authentication
   .configure(authentication({
-    idField: 'id'
+    idField: 'id',
+    header: 'X-Auth-Token'
   }))
   // Initialize a user service
   .use('/users', memory())
```

1. Set up the example server:

  ```bash
$ npm install
$ node example/app.js
Feathers authentication app started on 127.0.0.1:3030
```

1. In a new terminal session, create a new token using the default user:

  ```bash
$ curl -X POST \
-H 'Content-Type: application/json' \
-d '{ "email": "admin@feathersjs.com", "password": "admin" }' \
http://127.0.0.1:3030/auth/local
```

1. Use the token to view a restricted resource:

  ```bash
$ curl -X GET \
-H 'X-Auth-Token:$YOUR_TOKEN_FROM_ABOVE' \
http://127.0.0.1:3030/messages
```

  Despite using the header I configured the app to use this request fails for me.

mmwtsn:
#### Background

This ensures that if a custom server-side `header` value is set authentication still works. Currently on master requests fail if a custom `header` value is set to a string with capital letters. See the linked issue for additional background. If this patch is accepted I'll update the docs.

Fixes #218.

#### Test steps

1. Add an example `header` key with `"Example-Header"` to the example app.

1. Create a new token by making a `POST` request to `/auth/local` with the default user credentials.

1. Use that token with to make a request to `/messages`.

  - [ ] Verify that you are able to authenticate with a capitalized header (`Example-Header`).
  - [ ] Verify that you are able to authenticate with a lowercase header (`example-header`).

### DIFF
--- a/src/client/hooks.js
+++ b/src/client/hooks.js
@@ -13,7 +13,7 @@ export function populateHeader(options = {}) {
   return function(hook) {
     if (hook.params.token) {
       hook.params.headers = Object.assign({}, {
-        [options.header || 'Authorization']: hook.params.token
+        [options.header || 'authorization']: hook.params.token
       }, hook.params.headers);
     }
   };
--- a/src/middleware/express.js
+++ b/src/middleware/express.js
@@ -32,7 +32,8 @@ export function normalizeAuthToken(options = {}) {
   }
 
   return function(req, res, next) {
-    let token = req.headers[options.header];
+    // Normalize header capitalization the same way Node.js does
+    let token = req.headers[options.header.toLowerCase()];
 
     // Check the header for the token (preferred method)
     if (token) {
--- a/test/integration/rest.test.js
+++ b/test/integration/rest.test.js
@@ -12,6 +12,7 @@ describe('REST authentication', function() {
   let password = 'test';
   let settings = {
     idField: 'id',
+    header: 'X-Auth',
     token: {
       secret: 'feathers-rocks'
     }
@@ -259,7 +260,7 @@ describe('REST authentication', function() {
             method: 'GET',
             json: true,
             headers: {
-              Authorization: validToken
+              'X-Auth': validToken
             }
           };
         });


marshallswain:
I thought that the body-parser module automatically converted them to lowercase.  I think we definitely want this if it's not happening already.

joshuajabbour:
The body-parser module doesn't touch the headers.

marshallswain:
Well, see what I know? :)  Whichever module it is, I’ve not seen an uppercase header in node.

joshuajabbour:
Yeah, node itself does the lowercasing iirc. So yes, Feathers should def also do the same.

joshuajabbour:
Yep, node: https://nodejs.org/api/http.html#http_message_headers :)

ekryski:
@mmwtsn 👍 awesome! Would you be able to add a quick test in auth to make sure that headers are being lower cased?

mmwtsn:
@ekryski thanks! I pushed up a commit lowercasing the client-side header as discussed in https://github.com/feathersjs/feathers-authentication/issues/218#issuecomment-223716099 as well as a test for the capitalized header bug. Without `.toLowerCase()` the test fails.

I opened up [a pull request](feathersjs/feathers-docs/pull/184) to update the docs accordingly once this is merged.

daffl:
Thanks again! This has just been released in v0.7.8

PR #225, opened 2016-06-15
added support for hashing passwords when hook.data is an array

eblin:
feathers-sequelize allows you to bulk create (maybe other adapters do too?), in my case I was bulk creating predefined set of users (like admins, etc) and I noticed the auth was failing and later discovered the passwords were not being hashed at all.

We could always loop thru the array and call `.create()` individually, but *I think* this should be supported by the `hashPassword` hook, I certainly thought it was!

### DIFF
--- a/src/hooks/hash-password.js
+++ b/src/hooks/hash-password.js
@@ -4,8 +4,8 @@ import bcrypt from 'bcryptjs';
 
 const defaults = { passwordField: 'password' };
 
-export default function(options = {}){
-  return function(hook) {
+export default function (options = {}) {
+  return function (hook) {
     if (hook.type !== 'before') {
       throw new Error(`The 'hashPassword' hook should only be used as a 'before' hook.`);
     }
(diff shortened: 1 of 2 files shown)


marshallswain:
This looks good.   I absolutely agree that this should be possible.  Other official db adapters support bulk actions, too.  

Also FYI, it's likely that in an upcoming Feathers release that we'll be making bulk actions an opt-in feature per service.

daffl:
Makes sense. We don't want to add the lib folder though. Everything in that folder is generated and should not be changed or checked in.

eblin:
Ok thanks for the heads up,

Yes sorry about the lib folder commit, I thought those commits were not going to make it to the PR, anyways that was just needed for my build since npm does not run `prepublish` hooks when using your own forked modules (https://github.com/npm/npm/issues/3055), so anyone forking the library and wants to use it would have to compile manually.

Sorry about that! Feel free to ignore/clean up the last 2 commits

eblin:
I went ahead and removed the last 2 commits and moved them to a branch to keep the PR clean.

I can still use my version in my build just fine thanks to npm letting you point to a branch or commit when using your own modules.

daffl:
If we merge this PR and make a release you won't have to use your own branch anymore right?

ekryski:
👍  @eblin this is all good! Thanks for such a great PR!

eblin:
@daffl that's correct.
@ekryski no problem, thank you guys for great framework!

ekryski:
Released v0.7.9 with this fix in! 🍻 @eblin!

PR #229, opened 2016-07-03
Adds ability to limit queries unless authenticated and authorized

codingfriend1:
Often developers want to allow unauthenticated users or visitors to query data from a table but place restrictions on what type of data in the table is returned based on information in each row. This allows you to merge a restriction query into the query params to limit of the scope of what an unauthenticated user may search for.

Adds 3 hooks:

- verifyOrRestrict({ restrict: {approved: true} })
- populateOrRestrict({ restrict: {approved: true} })
- hasRoleOrRestrict({roles: ['admin'], restrict: {approved: true} })

NOTICE: Currently does not filter direct ids through the get method

### DIFF
--- a/example/app.js
+++ b/example/app.js
@@ -45,10 +45,19 @@ approvedMessageService.create({text: 'A million people walk into a Silicon Valle
 approvedMessageService.create({text: 'Nobody buys anything', approved: true}, {}, function(){});
 approvedMessageService.create({text: 'Bar declared massive success', approved: true}, {}, function(){});
 
+
 // Will merge this restriction with the query params
 var restriction = { restrict: {approved: true} };
 
 approvedMessageService.before({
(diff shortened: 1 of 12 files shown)


codingfriend1:
Should now restrict direct id queries. However if using memory it's necessary to use
```
all: [function(hook) {
      if(hook.id) {
        hook.id = parseInt(hook.id, 10);
      }
}]
```

to make the hook.id an integer.

ekryski:
@codingfriend1 thanks for the PR! I will take a look at this tomorrow or Friday. 🍻

ekryski:
Actually this is totally awesome @codingfriend1! :shipit: Would you mind doing a PR to add these hooks to the docs? http://docs.feathersjs.com/authorization/bundled-hooks.html

PR #231, opened 2016-07-05
RestrictToOwner now takes an array

sscaff1:
Basic use case:
- You have a model (let's call it a team model) with an array ids as admins
- You want to check who has permission to modify the team by checking the authenticated user against the admin array

If you guys don't think this type of functionality is appropriate in the bundled hooks, I understand. Just thought I would throw it out there since I just ran into a use case for this functionality in the project I'm currently working on.

### DIFF
--- a/src/hooks/restrict-to-owner.js
+++ b/src/hooks/restrict-to-owner.js
@@ -28,7 +28,7 @@ export default function(options = {}){
     }
 
     options = Object.assign({}, defaults, hook.app.get('auth'), options);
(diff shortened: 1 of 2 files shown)


ekryski:
Yes I think this is good! Thanks @sscaff1! We're going to be revisiting/putting up a proposal for normalizing permissions and probably moving the auth hooks to their own repo but this will definitely go in for now! Thanks for the PR! 😄

PR #232, opened 2016-07-07
Throws an error if restriction methods are used outside of a find or get hook

codingfriend1:
An extra safety check for the developer.


ekryski:
Thanks! 😄

PR #245, opened 2016-07-18
overwrite password check

MichaelErmer:
This is to allow custom password encryption strategies like:
```
'use strict';
const authentication = require('feathers-authentication');
const crypto = require('crypto');
    
module.exports = function() {
  const app = this;
    
  let config = app.get('auth');
    
  function verifyPassword(password, user, done) {
      var shasum=crypto.createHash("sha1");
      shasum.update(app.get("salt")+"&&&&&"+password);
      var p=shasum.digest("hex");
      return done(null, p===user.password);
  }
  config.local.verifyPassword = verifyPassword;
  app.set('auth', config);
};

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,8 @@ node_modules
 # Users Environment Variables
 .lock-wscript
 
+# Temp files
+*~
+
 lib/
 data.db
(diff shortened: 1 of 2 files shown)


ekryski:
Thanks for the PR @MichaelErmer! Although it's hidden you can actually already do this.

All you need to do is pass a function that implements the following signature `crypto.compare(password, hash, function(error, result) {` through `options.bcrypt`. Does that work for you?

MichaelErmer:
@ekryski I guess that will do it.

PR #252, opened 2016-07-31
Attempt to get token right away.

marshallswain:
This makes it so that we don’t have to wait for an async response in order to start making authenticated requests.


marshallswain:
@ekryski Any idea how to setup the token on socket.io?  This isn't doing the trick for some reason.

ekryski:
👍  :shipit: no need to worry about tests right now. I'll have to fix/re-write a bunch

PR #255, opened 2016-08-01
Don’t mix options when signing tokens

### LINKED ISSUE #254: 0.8: Token Service errors on token auth using client
On this line: https://github.com/feathersjs/feathers-authentication/blob/0.8/src/services/token.js#L144
```
jwt.sign(data, secret, options, (error, token) => {
```

If you pass in any of the following combinations, you get an error from the `jsonwebtoken` package:

`data.iss && options.issuer`
`data.sub && options.subject`
`data.exp && options.expiresIn`

marshallswain:
Fixes #254.  Allows the attributes on `data` to override the  options passed to `jwt.sign`.

### DIFF
--- a/src/services/token.js
+++ b/src/services/token.js
@@ -116,15 +116,22 @@ export class Service {
     const secret = this.options.secret;
     const options = {
       algorithm,
-      expiresIn,
       notBefore,
       audience,
-      issuer,
       jwtid,
-      subject,
       noTimestamp,
       header
     };
+
+    if (!data.iss) {
+      options.issuer = issuer;
+    }
+    if (!data.sub) {
+      options.subject = subject;
+    }
+    if (!data.exp) {
+      options.expiresIn = expiresIn;
+    }
 
     // const data = {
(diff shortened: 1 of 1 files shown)


src/services/token.js
```diff
@@ -116,16 +116,23 @@ export class Service {
...
       notBefore,
       audience,
-      issuer,
       jwtid,
-      subject,
       noTimestamp,
       header
     };
 
+    if (!data.iss) {
```
ekryski: Confused a bit about this one @marshallswain. Maybe I'm missing something. Any reason you want to pass those as `data`?

I was thinking any token options should come from `params.jwt` and due to the object deconstruction above this, those fields will just be `undefined` if they weren't passed along via `params.jwt` or aren't a default value.

marshallswain: They are already getting passed when you login with a token, which is
causing an error.

ekryski: Ah. Ok cool.

PR #256, opened 2016-08-01
Prevent redirect on successful auth/token login.

marshallswain:
Since the default value for successRedirect is null, we have to check for `null` to prevent auth/token logins from breaking.  Otherwise, they get redirected to `/auth/null`.

### DIFF
--- a/src/middleware/express/login-success.js
+++ b/src/middleware/express/login-success.js
@@ -4,7 +4,7 @@
 
 export default function loginSuccess(options = {}) {
   return function(req, res, next) {
-    if (options.successRedirect !== undefined) {      
+    if (options.successRedirect !== undefined && options.successRedirect !== null) {      
       return res.redirect(options.successRedirect);
     }
     


src/middleware/express/login-success.js
```diff
@@ -4,7 +4,7 @@
 
 export default function loginSuccess(options = {}) {
   return function(req, res, next) {
-    if (options.successRedirect !== undefined) {      
+    if (options.successRedirect !== undefined && options.successRedirect !== null) {
```
ekryski: It should probably just be `if (options.successRedirect) {...}`. I don't know why I didn't do that. Might have been a copy-paste remnant from how we were handling redirects pre 0.8. That makes more sense I think.

ekryski: I can fix this really quickly here @marshallswain unless you've already made the change I suggested.

marshallswain: Go for it.  I haven't gotten back to this, yet.

ekryski:
Fixed in 0.8 branch which is landing shortly.

PR #278, opened 2016-08-23
Manually supply an endpoint to the Client authenticate() method

mcnamee:
By simply providing the ability to specify the endpoint in the client authenticate method, it allows the ability for mobile users to authenticate via a native mobile Facebook SDK for example, and then post that Facebook access_token to Feathers to do it's thing via the FacebookTokenStrategy method.

eg. from react-native, I've just authenticated natively via the Facebook SDK, which has given me an access token:

```
app.authenticate({
   type: 'token',
   endpoint: '/auth/facebook',
   access_token: accessToken
});
```

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -52,7 +52,9 @@ export default function(opts = {}) {
       return getOptions.then(options => {
         let endPoint;
 
-        if (options.type === 'local') {
+        if (options.endpoint) {
+          endPoint = options.endpoint;
+        } else if (options.type === 'local') {
           endPoint = config.localEndpoint;
         } else if (options.type === 'token') {
           endPoint = config.tokenEndpoint;


marshallswain:
This looks good to me.  Do you mind making a PR for the docs and we can merge them both at the same time?

mcnamee:
No problem @marshallswain - I've just submitted a PR. Hope it's ok :)

mcnamee:
I've also added a `catch` for when an auth error occurs. Without this update, the script always continues as successful regardless of whether an error occurred. Now the client can show an error message etc.

daffl:
I think this looks great, thank you @mcnamee! Looks like JSHint is failing on CI at the moment (`src/client/index.js: line 54, col 8, Missing semicolon.`).

mcnamee:
@daffl - sorry for wasting your time, I should have tested sooner!

daffl:
No worries. Thanks a lot for the PR! Released in 0.7.10

PR #280, opened 2016-08-26
Fix: the data was not prepared correctly for creation/updation function

harish2704:
Before this patch, the `data` passed to `creation` or 'patch'  was some thing like ( eg: let provider = github )

```
{
  githubId: xxx,
  github: {
    email: xxx,
    gravatar_id: '',
    url: 'https://api.github.com/users/harish2704',
    html_url: 'https://github.com/harish2704',
  }
}
```

And this was causing problem in the case of `creating a new user` and `updating existing user` at the time of login

Please let me know if I am correct or not


ekryski:
@harish2704 Thanks for taking the time to put together a PR. We've just put out a new pre-release version of auth `v1.0.0-alpha` that was a major overhaul. This is now resolved in https://github.com/feathersjs/feathers-authentication-oauth2, which also makes it much easier to customize the response from the OAuth provider.

PR #283, opened 2016-08-30
Allow multiple username fields for local auth

sdbondi:
```javascript
"auth": {
  "local": {
   "usernameFields": ["username", "email"]
  }
}
```

If the request contains username or email (in that order) that will be used 
to locate the user.

`usernameField` still works as before. `usernameFields` is optional and will override `usernameField` if both are specified.

### DIFF
--- a/src/services/local.js
+++ b/src/services/local.js
@@ -32,7 +32,7 @@ export class Service {
 
       const params = {
         query: {
-          [field]: username
+          [field]: req.body[field],
         }
       };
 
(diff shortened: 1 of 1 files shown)


sdbondi:
Ok, mistake admitting time - this PR depends on a PR in `passport-local` being merged (https://github.com/jaredhanson/passport-local/pull/148) :disappointed:

ekryski:
See my comment on https://github.com/jaredhanson/passport-local/pull/148#issuecomment-261506180

Issue #286, opened 2016-09-08
hashPassword in patch


mcchrish:
This check https://github.com/feathersjs/feathers-authentication/blob/master/src/hooks/hash-password.js#L36-L39 is preventing patch methods if password is not defined in the data.

marshallswain:
Yeah.  `hashPassword` should just hash a password if one is available.  I made a PR. 

cc: @ekryski, @daffl

mcchrish:
An alternative for now is:
```js
const { hooks } = require('feathers-authentication');

// Your custom hashPassword hook
exports.myHashPassword = function(options) {
  const hashPassword = hooks.hashPassword(options);

  // Add any custom options
  return function(hook) {
    if(hook.data.password !== undefined) {
      // Call the hashPassword hook
      return hashPassword.call(this, hook);
    }
  };
}
```

Taken from http://docs.feathersjs.com/hooks/usage.html#customizing-built-in-hooks

PR #287, opened 2016-09-08
hashPassword fall-through if there's no password

marshallswain:
The user should have already run a `requireAuth` hook by this point. Fixes https://github.com/feathersjs/feathers-authentication/issues/286


daffl:
Can you fix the tests and then we can get a new version out?

PR #298, opened 2016-10-02
Allow to complete registration

bertho-zero:
Allow to complete registration with oauth2 login and a user hook:

```
hook => {
  if (hook.params.body && hook.params.body.user) {
    hook.data = { ...hook.data, ...hook.params.body.user };
  }
}
```

### DIFF
--- a/src/services/oauth2.js
+++ b/src/services/oauth2.js
@@ -69,7 +69,7 @@ export class OAuth2Service {
         debug(`Creating new user with ${options.provider}Id: ${profile.id}`);
 
         // No user found so we need to create one.
-        return app.service(options.userService).create(data).then(user => {
+        return app.service(options.userService).create(data, {body: req.body}).then(user => {
           debug(`Created new user: ${user[options.idField]}`);
           // TODO (EK): Handle paginated services?
           return done(null, user);


ekryski:
@bertho-zero thanks for the PR! Can you elaborate on what the use case is for this? I think I might be missing something. 🍻

bertho-zero:
I created a branch for illustrate it:

When I log me first time, with FB for example, the user service is requested to create a user, here I have [an error](https://github.com/bertho-zero/react-redux-universal-hot-example/blob/feathers-auth/api/services/users/hooks.js#L15-L16) because it lacks the necessary data for registration.
From there [I redirected](https://github.com/bertho-zero/react-redux-universal-hot-example/blob/feathers-auth/src/containers/Login/Login.js#L29-L34) to the registration completion with oauth data and I [retry](https://github.com/bertho-zero/react-redux-universal-hot-example/blob/feathers-auth/src/containers/Register/Register.js#L22-L24) with the missing information.

But I found how to do this would otherwise (https://github.com/bertho-zero/react-redux-universal-hot-example/commit/46bcd1ea044db8e88ecd6aa47f7cc151aedaef72)

PR #304, opened 2016-10-06
0.8 - OAuth fixes

marshallswain:
~~This fixes a couple of problems that were occurring in the 1st half of the OAuth flow (up until showing the GitHub auth page)~~  I think the entire OAuth flow should work for every supported provider, now.  

1. The feathers provider plugin name was overwriting the OAuth provider name.
2. If the callbackURL didn't have a leading `/`, Passport would mess it up. (includes tests)

See commit messages for more details.

### DIFF
--- a/src/services/oauth2.js
+++ b/src/services/oauth2.js
@@ -38,6 +38,7 @@ export class OAuth2Service {
   updateUser(user, data) {
     const idField = this.options.user.idField;
     const id = user[idField];
+    const userService = typeof this._userService === 'string' ? this.app.service(this._userService) : this._userService;
 
     debug(`Updating user: ${id}`);
 
@@ -45,15 +46,16 @@ export class OAuth2Service {
     data = merge({}, user, data);
 
     // TODO (EK): Handle paginated services?
-    return this._userService.patch(id, data, { oauth: true });
+    return userService.patch(id, data, { oauth: true });
   }
 
   createUser(data) {
     const provider = this.options.provider;
     const id = data[`${provider}Id`];
+    const userService = typeof this._userService === 'string' ? this.app.service(this._userService) : this._userService;
     debug(`Creating new user with ${provider}Id: ${id}`);
 
-    return this._userService.create(data, { oauth: true });
+    return userService.create(data, { oauth: true });
   }
 
   verify(req, accessToken, refreshToken, profile, done) {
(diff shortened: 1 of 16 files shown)


marshallswain:
I fixed a few more things, but it's still not quite finished.  The OAuth flow completes correctly, now, but the response needs to be fixed, still.   It's currently returning the JSON (includes the token and user fields) in the response.  We need to implement the success callback, which is probably something that will need to be configured for each OAuth provider.  I suppose we could still use `/auth/success` as the default.

@daffl, @ekryski, @corymsmith

src/services/oauth2.js
```diff
@@ -36,22 +38,24 @@ export class OAuth2Service {
   updateUser(user, data) {
     const idField = this.options.user.idField;
     const id = user[idField];
+    const userService = typeof this._userService === 'string' ? this.app.service(this._userService) : this._userService;
```
ekryski: @marshallswain rather than duplicating this in each method can we DRY it up and put this in the `setup` like I have here: https://github.com/feathersjs/feathers-authentication/blob/0.8/src/services/oauth2.js#L176-L177?

ekryski:
Other than my one comment, looks good. :shipit:

We have [the default success handler](https://github.com/feathersjs/feathers-authentication/blob/0.8/src/services/oauth2.js#L229) (which is just redirect middleware) so you can set `successRedirect` for each OAuth provider you register and it will handle it. Alternatively you can set your own `successHandler` if you want to register your own middleware after OAuth.

So I think we are pretty well covered there but maybe if we are using the default successhandler and don't have a successRedirect we should set a default one or throw an error telling the developer they need to set one for OAuth. I personally prefer the latter.

src/services/oauth2.js
marshallswain: That's how it was originally setup, but if you do something like this:

```
app.configure(authentication);
app.use('/api/users', service(options));
```
The users service hasn't been setup by the time the auth plugin's `setup` method runs, it's still `undefined` until after the users service gets setup.  I made it more DRY by putting the lookup into a class method in that last commit.

marshallswain:
@ekryski I committed a line for the successHandler.  Please check to make sure I understood you correctly.

src/services/oauth2.js
ekryski: Hmm. I don't remember that being an issue but you could be right that order might matter. In any case this is totally good for now. I might double check that order matters when it is in the `setup` method but it's pretty low priority right now. 👍

ekryski:
yup looks good homie. :shipit:

PR #305, opened 2016-10-09
First cut for authentication middleware

daffl:
All original tests are still passing. Update of the existing codebase will happen in another PR.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,0 +1,59 @@
+import merge from 'lodash.merge';
+
+// Options that apply to any provider
+export const defaults = {
+  header: 'Authorization',
+  setupMiddleware: true, // optional - to setup middleware yourself set to false.
+  cookie: { // Used for redirects, server side rendering and OAuth
+    enabled: false, // Set to true to enable all cookies
+    name: 'feathers-jwt',
+    httpOnly: true,
+    maxAge: '1d',
+    secure: true
+  },
+  jwt: {
+    issuer: 'feathers',
+    algorithm: 'HS256',
+    expiresIn: '1d'
+  },
+  token: {
+    name: 'token', // optional
+    service: '/auth/token', // optional string or Service
+    subject: 'auth', // optional
+    issuer: 'feathers', // optional
+    algorithm: 'HS256', // optional
+    expiresIn: '1d', // optional
+    secret: null, // required
+    successRedirect: null, // optional - no default. If set the default success handler will redirect to location
+    failureRedirect: null, // optional - no default. If set the default success handler will redirect to location
+    successHandler: null // optional - a middleware to handle things once authentication succeeds
+  },
+  local: {
+    service: '/auth/local', // optional string or Service
+    successRedirect: null, // optional - no default. If set the default success handler will redirect to location
+    failureRedirect: null, // optional - no default. If set the default success handler will redirect to location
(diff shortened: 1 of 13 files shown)


src/options.js
```diff
@@ -0,0 +1,59 @@
...
+  },
+  jwt: {
+    issuer: 'feathers',
+    algorithm: 'HS256',
+    expiresIn: '1d'
+  },
+  token: {
+    name: 'token', // optional
+    service: '/auth/token', // optional string or Service
+    subject: 'auth', // optional
```
ekryski: Good call moving this into a separate file and setting up `jwt` namespace. I guess we'll de-dupe these options later?

src/services/authentication.js
```diff
@@ -0,0 +1,19 @@
...
+      return Promise.reject(new Error(`External ${params.provider} requests need to run through an authentication provider`));
+    }
+
+    return this.authentication.create(data.payload);
+  }
+
+  remove(id, params) {
+    const token = id !== null ? id : params.token;
+
+    return this.authentication.verify({ token });
```
ekryski: Doesn't this need to be `verifyToken`?

src/token/from-request.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  return function fromRequest(data) {
+    if (typeof data === 'object' && data.headers) {
+      const req = data;
+
+      debug('Parsing token from request');
+
+      // Normalize header capitalization the same way Node.js does
+      let token = req.headers && req.headers[header.toLowerCase()];
```
ekryski: Do we want to support body and query string as well or just header? If it's not going to cause us problems maybe we should (at least body). I personally use querystring a lot in dev mode because I'm lazy but really we shouldn't support that officially.

src/token/populate-user.js
```diff
@@ -0,0 +1,35 @@
...
+export default function(options) {
+  const app = this;
+  const { user } = options;
+
+  if(!user.service) {
+    throw new Error(`'user.service' needs to be set in authentication options`);
+  }
+
+  return function populateUser(data) {
+    const service = app.service(user.service);
```
ekryski: Possibly we want to support passing an actual service as well, instead of just the path name.

src/options.js
daffl: Yep, we'll clean it up later. De-duping and the `local` and `token` sections might change as well.

src/token/from-request.js
daffl: How is `body` being used? I couldn't think of a case where it made much sense (even might cause issues with `token` being a legitimate property when calling `create`, `update` and `patch`).

Do you have use cases where query string is easier than setting the header? We can add it back in but I don't think it was documented so far with more than a brief mention. Removing it would be less to document and test and we don't risk promoting not a best practise.

marshallswain: I would vote to nix both and only potentially add support for body if it is requested after release.

ekryski: I agree actually. You're totally right @daffl we shouldn't support body. Querystring might still be a thing actually if we are supporting short lived JWTs for things like pw reset however, let's just go with this for now. I'm going to tinker with something regarding the password reset tokens, possibly by only allowing querystring verification if the token is an `authentication` subject.

ekryski:
:shipit:

PR #306, opened 2016-10-10
Change cookie default options.

marshallswain:
Cookies are still off by default, but when enabled, the default expiry will match the token life.  Setting a cookie expiry or maxAge will override the default setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "bcryptjs": "^2.3.0",
     "cookie-parser": "^1.4.3",
     "debug": "^2.2.0",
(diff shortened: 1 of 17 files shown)


daffl:
Looks like what we discussed. I let @ekryski double check and then we can :shipit:

ekryski:
:shipit: 

I was thinking we should have 2 different cookies, one for OAuth and one for SSR that you can enable independently but that shouldn't block this at all. That can be added in another PR.

marshallswain:
I merged this PR instead: https://github.com/feathersjs/feathers-authentication/pull/308

PR #311, opened 2016-10-13
Only set app token if jwt not expired.

marshallswain:
Modifies the getJWT function to return undefined if the token has expired.  Since we run `getJWT` right away and then set the token with `app.set('token, jwt)`, this gives you a quick way of knowing if the user is (more than likely) authenticated without having to make a request.

### DIFF
--- a/src/client/utils.js
+++ b/src/client/utils.js
@@ -1,3 +1,5 @@
+import decode from 'jwt-decode';
+
 // Returns a promise that resolves when the socket is connected
 export function connected(app) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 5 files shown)


src/client/utils.js
```diff
@@ -75,13 +77,13 @@ export function clearCookie(name) {
 // Tries the JWT from the given key either from a storage or the cookie
 export function getJWT(tokenKey, cookieKey, storage) {
   return Promise.resolve(storage.getItem(tokenKey)).then(jwt => {
-    const cookieToken = getCookie(cookieKey);
-
-    if (cookieToken) {
-      return cookieToken;
+    let token = jwt || getCookie(cookieKey);
+    if (token) {
+      if (decode(token).exp * 1000 < new Date().getTime()) {
```
ekryski: can we wrap this in a separate `isExpired` function or something? Might be handy to use outside of just the `getJWT` call on the client.

marshallswain:
Closing in favor of https://github.com/feathersjs/feathers-authentication/pull/314

PR #314, opened 2016-10-14
0.8 client decode jwt

marshallswain:
This adds a couple of util methods:
- `payloadIsValid` returns true if the payload hasn't expired.
- `decodeJWT` takes a token and returns the payload if it was valid.

`getJWT` has also been updated to only return the token if the token payload is valid.

Finally, `decodeJWT` is added to the `app` object, so you can do `app.decodeJWT(app.get('token'))` to get the token payload.  It's the same as the `decodeJWT` util, so it only returns a payload if it was valid.

Includes tests.  Will need docs before release:


daffl:
Should we make it the same as on the server where we now have

```js
app.authentication.options // authOptions
app.authentication.verifyJWT // decode + verify
app.authentication.getJWT // get token
```

For the server `app.authentication` is an instance of this class: https://github.com/feathersjs/feathers-authentication/blob/0.8/src/base.js

marshallswain:
Definitely.  I'll update it tomorrow.

marshallswain:
@feathersjs/core-team do we want to keep the logout method where it is at `app.logout()`, or move it to `app.authentication.logout()`?

ekryski:
hmm. I think keep it where it's at. We have `app.authenticate` as well.

slajax:
I like having alias easily accesssible on app. That said, its more important to me that it is consistent if there is a client / server implementation.

marshallswain:
Ok, so we'll have.
```
app.authenticate()
app.logout()
app.authentication.options // authOptions
app.authentication.verifyJWT // decode + verify
app.authentication.getJWT // get token
```

slajax:
:+thumbsup

ekryski:
Ya let's go with that. Once you get those changes in @marshallswain ping one of us for a quick ship.

marshallswain:
Will do.  Also, I'm making both `verifyJWT` and `getJWT` sync, which won't match the server API (where both return promises), but I think promises will unnecessarily complicate things.  Let me know if you feel different.

marshallswain:
@ekryski we're not returning the user in the authenticate requests anymore, right?  So this stuff should be able to come out: https://github.com/feathersjs/feathers-authentication/blob/0.8-client-decode-jwt/src/client/index.js#L54

ekryski:
@marshallswain correcto. Will have to be a separate request by the client explicitly because you may not be authenticating a user. Could be a device, organization, toaster, whatever...

marshallswain:
Other than what we discussed above, the only additional change is that verifyJWT now allows passing an object with a token property on it, which matches the server.  This should be ready to ship unless anybody finds any issues.  If not, please pull the trigger and release.  I'm retiring for the night. 🌙 😴

.travis.yml
```diff
@@ -3,3 +3,5 @@ node_js:
   - 'node'
   - 'iojs'
   - '0.12'
+code_climate:
+  repo_token: 02ea69f85221efd3051a19e882c87cebc07522bf321374186bb6116301569e5e
```
ekryski: This is slightly different to get it working but I will update.

ekryski:
Nice! You pulled out the permissions stuff too. :shipit:

PR #317, opened 2016-10-17
client: use Authentication class, make `getJWT` and `verifyJWT` async

marshallswain:
This modifies both `getJWT` and `verifyJWT` to return promises.  `getJWT` has also been updated to check the store for the token again to support environments that don’t load cookies immediately.

Also renames the `getJWT` util function to `retrieveJWT` to avoid ambiguity.

### DIFF
--- a/src/client/authentication.js
+++ b/src/client/authentication.js
@@ -13,10 +13,23 @@ export default class Authentication {
   constructor (app, options) {
     this.options = options;
     this.app = app;
+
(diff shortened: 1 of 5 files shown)


daffl:
Didn't we want to create an `Authentication` class for this (we can always map some of its methods directly to the `app` object)? I can do that, too when wrapping this up though.

marshallswain:
@daffl I've updated it to use a class.

PR #324, opened 2016-10-23
Socket.io authentication tests and login logout event

daffl:
This pull request adds further tests for Socket.io authentication and implements login and logout events for it. Now it is possible to do this:

```js
app.on('login', function(authResult, info) {
  // authResult -> result from app.authentication.authenticate
  // info -> { socket, connection, provider }
});

app.on('logout', function(authResult, info) {
  // authResult -> result from app.authentication.authenticate
  // info -> { socket, connection, provider }
});
```

Getting the `info` object is useful so that we can e.g. keep the connection up to date with the latest user information (or let people do other things outside of the abstracted service flow).

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -5,7 +5,7 @@ import middlewares from './middleware/authentication';
 const debug = Debug('feathers-authentication:authentication:base');
 
 // A basic Authentication class that allows to create and verify JWTs
-// and also run through
+// and also run through a token authentication chain
 export default class Authentication {
   constructor(app, options) {
     this.options = options;
@@ -18,7 +18,7 @@ export default class Authentication {
     this._middleware.isInitial = true;
   }
 
-  // Register on or more handlers for the JWT verification chain
+  // Register one or more handlers for the JWT verification chain
   use(... middleware) {
     // Reset the default middleware chain
     if(this._middleware.isInitial) {
--- a/src/base.js
+++ b/src/base.js
@@ -4,6 +4,8 @@ import middlewares from './middleware/authentication';
 
 const debug = Debug('feathers-authentication:authentication:base');
 
+// A basic Authentication class that allows to create and verify JWTs
+// and also run through
 export default class Authentication {
   constructor(app, options) {
     this.options = options;
@@ -16,6 +18,7 @@ export default class Authentication {
     this._middleware.isInitial = true;
   }
 
+  // Register on or more handlers for the JWT verification chain
   use(... middleware) {
     // Reset the default middleware chain
     if(this._middleware.isInitial) {
@@ -31,6 +34,7 @@ export default class Authentication {
     return this;
   }
 
+  // Run the JWT verification chain against data
   authenticate(data) {
     let promise = Promise.resolve(data);
 
@@ -46,6 +50,8 @@ export default class Authentication {
     return promise;
   }
 
+  // Returns a { token } object either from a string,
+  // an HTTP request object or another object with a `.token` property
   getJWT(data) {
     const { header } = this.options;
 
@@ -70,7 +76,7 @@ export default class Authentication {
         debug('Token found in header', token);
       }
 
-      return Promise.resolve({ token, req });
+      return Promise.resolve({ token });
     }
     else if (typeof data === 'object' && data.token) {
       return Promise.resolve({ token: data.token });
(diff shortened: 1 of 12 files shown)


src/base.js
```diff
@@ -4,6 +4,8 @@ import middlewares from './middleware/authentication';
 
 const debug = Debug('feathers-authentication:authentication:base');
 
+// A basic Authentication class that allows to create and verify JWTs
+// and also run through
```
marshallswain: // and also run through what?

src/base.js
```diff
@@ -16,6 +18,7 @@ export default class Authentication {
     this._middleware.isInitial = true;
   }
 
+  // Register on or more handlers for the JWT verification chain
```
marshallswain: on -> one

src/middleware/socket/handler.js
```diff
@@ -28,12 +15,25 @@ export default function setupSocketHandler(app, options, {
...
+          const { token } = result;
+          const connection = feathersParams(socket);
+
           debug(`Successfully authenticated socket with token`, token);
 
-          feathersParams(socket).token = token;
+          // Add the token to the connection so that it shows up as `params.token`
+          connection.token = token;
+
+          app.emit('login', result, {
```
marshallswain: How do you filter this?  Is there an `app.filter()` method?

marshallswain: Also, these events are already being emitted in the service, here: https://github.com/feathersjs/feathers-authentication/blob/socketio-auth/src/service.js#L17

Would it be better to only emit at only one of these?

marshallswain: Ok, so to emit the logout message to a single user, would we do this:
```
app.on('logout', (data, {socket}) => {
  socket.emit('logout', data);
});
```

ekryski: @daffl so you don't want to use `app.service('authentication').on('login')` and `app.service('authentication').on('logout')`? Is that because they don't have the provider context?

marshallswain: I only ask because it's what is wanted for bitcentive.  Will that send to only the authenticating or logging out connection if I do that?

ekryski: @daffl sounds good. Hopefully, because it's not on a service people won't be confused about it being dispatched outside of the server. I agree with you, I don't think they should be available to a client.

test/integration/socketio.test.js
```diff
@@ -25,54 +25,83 @@ describe('Socket.io authentication', function() {
...
+          id: 'laundry',
+          description: 'You have to do laundry',
+          user: { id: 0, name: 'Tester' }
+        });
+        done();
+      });
+    });
+  });
+
+  it('app login and logout events', done => {
```
daffl: That test was indeed doing too much, so I split it up. I guess the block of

```js
        expect(result.token).to.exist;
        expect(result.payload).to.exist;
        expect(result.payload.userId).to.equal(0);
        expect(result.authenticated).to.be.ok;
        expect(result.user).to.deep.equal({ id: 0, name: 'Tester' });

        expect(info.provider).to.equal('socketio');
        expect(info.socket).to.exist;
```

Technically still counts as one assertion since we are verifying the single object we are getting back, right?

ekryski: I typically still break them up into one `expect` per `it` block mostly because if you change the result coming back or remove an attribute it's easier to determine which test needs to change.

However, that only applies if the setup cost is pretty negligible, which it might not be in this particular case.

src/middleware/socket/handler.js
ekryski: @marshallswain we had talked about that. I mean the client will get an ack for the `authenticate` event it sends to the server, and same for logout so we don't really need to emit those events to the client as well.

ekryski:
So I think other than those typos @marshallswain pointed out this is good to 🚢

test/integration/socketio.test.js
```diff
@@ -14,65 +14,155 @@ describe('Socket.io authentication', function() {
...
   after(function() {
     this.server.close();
   });
 
   it('returns not authenticated error for protected endpoint', done => {
+    socket.emit('todos::get', 'laundry', e => {
+      delete e.stack;
+      delete e.type;
+
+      expect(e).to.deep.equal({
```
ekryski: Minor comment. I usually just check for the error code. That way if we ever change any of that other stuff we don't need to update a ton of tests.

daffl: Word. I'll update that as well.

PR #332, opened 2016-10-26
restrictToOwner -Fix check for methodNotAllowed

### LINKED ISSUE #319: restrict-to-owner throws error when user id is 0
I was doing some testing and created a test user in `feathers-memory` like this:
```
userService.create({
  email: 'test@example.com',
  password: 'password',
}).then((value) => {
  console.log('created user:', value);
});
```

Then when I tried to authenticate with `auth/local` I got the error: `The 'restrictToOwner' hook should only be used on the 'get', 'update', 'patch' and 'remove' service methods.` I knew that I was only using the hook on those 4 service methods so I dug into the `restrict-to-owner` code and found this:
```
if (!hook.id) {
  throw new _feathersErrors2.default.MethodNotAllowed('The \'restrictToOwner\' hook should only be used on the \'get\', \'update\', \'patch\' and \'remove\' service methods.');
}
```

The `!hook.id` check fails when the id is 0 (or any other false value). This check should probably be more careful. (I worked around it by giving a proper id)

aengushearne:
Hi!

methodNotAllowed was being checked as "if (!hook.id)", leading to erroneous or misleading error messages.
Changed to check against hook.method 

This also fixes #319

### DIFF
--- a/src/hooks/restrict-to-owner.js
+++ b/src/hooks/restrict-to-owner.js
@@ -13,7 +13,7 @@ export default function(options = {}){
     }
     
     // Check hook is being called on an allowable method
-    if (!hook.method === 'get' || !hook.method === 'update' || !hook.method === 'patch' || !hook.method === 'remove') {
+    if (!(hook.method === 'get' || hook.method === 'update' || hook.method === 'patch' || hook.method === 'remove')) { {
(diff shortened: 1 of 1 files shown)


daffl:
Awesome, thank you! Looks like the test (around https://github.com/feathersjs/feathers-authentication/blob/master/test/src/hooks/restrict-to-roles.test.js#L160) needs to be updated accordingly.

PR #333, opened 2016-10-27
0.8 client oauth utils

marshallswain:
Adds utilities for managing auth popups

This adds an `openLoginPopup` util function that takes a URL and opens an OAuth window (or whatever URL you give it) in a popup.  It also creates a global `authAgent` EventEmitter on the main page, which can be used to receive data from the popup after OAuth redirects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "bcryptjs": "^2.3.0",
     "cookie-parser": "^1.4.3",
     "debug": "^2.2.0",
+    "events": "^1.1.1",
     "feathers-commons": "^0.7.5",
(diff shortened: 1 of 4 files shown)


marshallswain:
Dang.  Those tests pass locally.  I didn't touch the failing ones.  This is probably due to adding the -jsdom stuff to the mocha tests for the new client stuff.  What I should probably do, instead, is setup the `feathers-authentication-client` repo to start with version `0.8` and move this stuff over to that repo.

marshallswain:
This has been moved into its own repo: https://github.com/feathersjs/feathers-authentication-popups

PR #336, opened 2016-10-29
1.0 Pre-release

ekryski:
### Summary

This is a pretty big overhaul and now utilizes passport for the core authentication. This allows us to use **any** passport strategy from any location: hooks, express middleware, sockets. It also future proofs us so that all we need to do is adapt any other framework or transport to match the expected Passport `request` object.
### Still to do:
- [x] Fix tests
- [x] Wrap up integration tests
- [ ] Fix Auth Client to be consistent
- [ ] Update docs
  - [x] Highlight new features
  - [x] Migration guide
  - [ ] Highlight data flow to create a JWT
  - [ ] Document public API/Options
  - [x] Document basic usage
  - [x] Document customizing JWT payload
  - [x] Document verifying a JWT payload
  - [ ] Document not sending a JWT
  - [ ] Document refresh tokens
  - [x] Document expected request object (writing a custom adapter)
- [ ] Show examples. These likely should just be guides on Medium with gists
  - [x] Username/Password auth (in [feathers-authentication-local](https://github.com/feathersjs/feathers-authentication-local/tree/master/example))
  - [ ] Basic auth
  - [x] JWT access token (in [feathers-authentication-jwt](https://github.com/feathersjs/feathers-authentication-jwt/tree/master/example))
  - [ ] Refresh tokens
  - [ ] OAuth1 (use twitter as example)
  - [x] OAuth2 (in [feathers-authentication-oauth2](https://github.com/feathersjs/feathers-authentication-oauth2/tree/master/example))
### Stretch Goals
- [ ] Add wrapper plugins for common authentication strategies that are are more feathers specific:
  - [ ] OAuth1
  - [x] [OAuth2](https://github.com/feathersjs/feathers-authentication-oauth2)
  - [x] [Local Auth](https://github.com/feathersjs/feathers-authentication-local)
  - [x] [JWT](https://github.com/feathersjs/feathers-authentication-jwt)

_These should remove a bunch of the passport boilerplate around setting up middleware routes, looking up the user, verifying password, and populating the user. They should pull from the main app config object. This should make for an easier transition for people as they generally can just be a one liner._ Example usage:

``` js
const auth = require('feathers-authentication');
const permissions = require('feathers-permissions');
const local = require('feathers-authentication-local');
const jwt = require('feathers-authentication-jwt');
const oauth2 = require('feathers-authentication-oauth2');
const FacebookStrategy = require('passport-facebook').Strategy;
const memory = require('feathers-memory');

app.configure(auth())
  .configure(local())
[...]

PR #340, opened 2016-11-03
Authenticate with custom endpoint over websockets

alanpog:
Same motivation as #278, but using websockets instead.  It was confusing that I could manually supply an endpoint using rest and it would be ignored if going over sockets.

This PR aims to do for websockets what #278 did for the rest provider, i.e., to allow the client to `authenticate` with a custom endpoint such as `auth/facebook`.

### DIFF
--- a/src/middleware/sockets.js
+++ b/src/middleware/sockets.js
@@ -46,7 +46,10 @@ function setupSocketHandler(feathersParams, provider, emit, app, options) {
 
         params.req.body = data;
 
-        app.service(options.localEndpoint).create(data, params).then(response => {
+        // If an endPoint was provided in the request, use it.
+        const endPoint = data.endpoint ? data.endpoint : options.localEndpoint;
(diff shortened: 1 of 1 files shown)


alanpog:
Is this PR made obsolete by #336 ?

ekryski:
Yeah but I wanted to get the docs and the client in order before I closed it so that I could explain why and how you can do what you want to do 😄

alanpog:
So let's keep this open as a reminder then, I'll be very grateful to be explained how to do this in 1.0. Thanks.

ekryski:
We haven't created a guide for it so would gladly take a PR to the docs for one, but you can easily change up the default path/endpoint on any of the authentication plugins that support it. All of the options are outlined in the docs.

For example, for OAuth2: https://docs.feathersjs.com/api/authentication/oauth2.html#default-options

Issue #348, opened 2016-11-23
[1.0] The entity on the socket should pull from the strategy options.


ekryski:
### Steps to reproduce

Set up a strategy with an entity that is different than `user`, say `organization`.

### Expected behavior
The socket object should have `socket.feathers.organization` and hooks should have `hook.params.organization`.

### Actual behavior
It's always `socket.feathers.user`.

### System configuration

Tell us about the applicable parts of your setup.

**Module versions** : v1.0.0-beta

PR #349, opened 2016-11-23
A bunch of bug fixes

### LINKED ISSUE #346: [1.0] Only the first failure is returned on auth failure when chaining multiple strategies
### Steps to reproduce

Register mulitple strategies like so and send a bad username and password.

```js
  before: {
    create: [
      auth.hooks.authenticate(['jwt', 'local'])
    ]
  }
});
```

### Expected behavior
It should return all failures or return the one for the strategy you were attempting to authenticate with.

### Actual behavior
If you have registered multiple strategies to be checked against, it only returns the first one that fails . This is very misleading as to what the actual error is that caused auth failure.

In the above example it if you provided a bad username and password auth would fail but it would return the failure from the `jwt` strategy, which would be that the token is missing or something.

### System configuration

Tell us about the applicable parts of your setup.

**Module versions**: feathers-authentication@1.0.0-beta

**NodeJS version**: all

**Operating System**: all

**Browser Version**: all

**React Native Version**: all

**Module Loader**: all

ekryski:
### Summary

- Now checks strategy options on sockets instead of the global options. #348 
- Now supporting entityId for JWT payload.
- Only authenticates against passed strategy. This results in more accurate errors and faster authentication time. Also errors if an unsupported strategy is provided. Closes #346.

### Other Information

Dependent on some fixes to

- feathers-authentication-jwt
- feathers-authentication-local
- feathers-authentication-oauth1
- feathers-authentication-oauth2
- feathers-authentication-client

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,8 +147,7 @@ app.service('authentication').hooks({
   before: {
     create: [
       // You can chain multiple strategies
-      auth.hooks.authenticate(['jwt', 'local']),
-      customizeJWTPayload()
+      auth.hooks.authenticate(['jwt', 'local'])
     ],
     remove: [
       auth.hooks.authenticate('jwt')
--- a/docs/migrating.md
+++ b/docs/migrating.md
@@ -91,29 +91,13 @@ app.use('/users', memory())
       Strategy: FacebookStrategy
     }));
 
-// This hook customizes your payload.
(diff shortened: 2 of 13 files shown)


ekryski:
This will go out as a 1.0.0-beta-1 release.

PR #353, opened 2016-11-25
fixing redirect middleware

ekryski:
### Summary

Express redirect middleware were not actually redirecting because of 2 things:

- the `hook` object lives on `res.hook` not `req.hook`
- if the hook object is previously modified the modifications weren't passed along. Only modified `hook.data` gets passed along to express middleware.


ekryski:
released in 1.0.0-beta-2

PR #362, opened 2016-12-07
Remove additional deprecation notice

porsager:
I think you might have forgotten to remove this deprecation notice and self referring link ;)


ekryski:
@marshallswain must have forgotten to remove it when he accidentally committed it. Thanks @porsager!

marshallswain:
Heh. Thanks! I guess the request didn't go through when I made my commit to remove it.

PR #368, opened 2016-12-13
Clarify that the authenticate hook is required.

marshallswain:
The "Hooks" heading wasn't enough to catch my attention that the authenticate hook is required.  It took me a while to figure out why my JWT didn't have any entity data in it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,9 @@ This module contains:
 5. Express middleware
 6. A [Passport](http://passportjs.org/) adapter 
 
-### Hooks
+### Required Hooks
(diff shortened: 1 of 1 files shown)


daffl:
Before merging lets finish the discussion in https://github.com/feathersjs/feathers-authentication-local/issues/4#issuecomment-266896570

ekryski:
I think this is probably a good call. I also want to remove the code climate shit that was added. It's making the repo look like shit because I have a couple TODOs in there and because the tests have "similar code".

PR #369, opened 2016-12-14
Logout should always give a response.

marshallswain:
I'm not sure what the correct way to handle this is, but this fixes https://github.com/feathersjs/feathers-authentication-client/issues/10.


daffl:
Great catch! Should this resolve with an empty object or throw an error?

marshallswain:
The goal is to end up without an authenticated user, which USB what you end up with even if you were never logged in.  So I don't think it should throw an error.  That's why I opted for the empty object.

ekryski:
Ya nice catch! Looks good. :shipit:

PR #376, opened 2016-12-16
add optional option to auth hook

marbemac:
It's really great that the new authentication populates the user. However, often-times it's nice to know if the user is authenticated, and have access to that data, but the authentication itself is not required. What do you guys think of a simple optional boolean exposed when using the hook?

As an example, we use the presence of an authenticated user to filter the properties returned to the client in find and get requests, but it is not required that the request be authenticated.

```js
exports.before = {
  all: [],
  find: [
    auth.hooks.authenticate(['local', 'jwt'], {optional: true}),
  ],
  get: [
    auth.hooks.authenticate(['local', 'jwt'], {optional: true}),
  ],
  create: [],
  update: [
    auth.hooks.authenticate(['local', 'jwt']),
  ],
  patch: [
    auth.hooks.authenticate(['local', 'jwt']),
  ],
  remove: [
    auth.hooks.authenticate(['local', 'jwt']),
  ],
};
```

### DIFF
--- a/src/hooks/authenticate.js
+++ b/src/hooks/authenticate.js
@@ -38,7 +38,7 @@ export default function authenticate (strategy, options = {}) {
     debug(`Attempting to authenticate using ${strategy} strategy with options`, options);
 
     return app.authenticate(strategy, options)(request).then((result = {}) => {
-      if (result.fail) {
+      if (result.fail && !options.optional) {
         // TODO (EK): Reject with something...
         // You get back result.challenge and result.status
         if (options.failureRedirect) {


ekryski:
Hey @marbemac! Thanks for the PR. Unfortunately, this would have bigger implications than just changing the hook. You would also need to change the socket and express middleware implementation to support the same thing.

With that said, auth can fail for all sorts of reasons (including user lookup) so I don't think this is a good idea. If you are looking for the functionality you can always implement your own strategy or custom verifier (see any of the auth plugin docs) or you could likely [wrap the hook](https://docs.feathersjs.com/hooks/usage.html#customizing-built-in-hooks), catch the error, and ignore it.

Couple ways you could support what you want I think. Hope you don't mind if I close this. Thanks again for taking the time! 😄

PR #380, opened 2016-12-18
Typo 'cookie.enable' should be 'cookie.enabled'

whollacsek:
cc feathersjs/feathers-authentication-oauth1#4


marshallswain:
Ah.  Thank you for helping clean up.  :)

PR #414, opened 2017-02-06
Prevent setCookie from mutating authOptions

adrien-k:
setCookie is mutating `authOptions`, setting a constant value for `expires` attribute which is then re-used on every further `setCookie` calls. 

After a while, cookies are sent already-expired...this PR should fix it (is it a proper way to do it ?).

Thanks for this awesome framework !


daffl:
Thank you for the fix and for the test! We'll release it shortly.

PR #419, opened 2017-02-13
Refresh token must have a user ID

francisco-sanchez-molina:


ekryski:
@francisco-sanchez-molina thanks for you the PR! However, this is breaking the tests. Any chance you can update with more details as to what this is fixing and also fix the tests and/or add new ones to cover the use case?

francisco-sanchez-molina:
Hi ekryski! Thank you for reviewing my PR. The test that fails is not related to my change, and it fails randomly :/. 

When you create a token, feathers return a jwt with user id inside payload at first level (payload:
 {_id:123}), but when refresh a token user id is located in a sub-level (payload: {data: {_id:123}}):

pasted image at 2017_04_11 09_10 am

And this broke some hooks that try to get user id from token

ekryski:
Yes I think this is fine. Thanks for updating the PR. I see that the tests are only failing on Node v0.12.0. Since that is no longer a supported version this should be good to be merged. :shipit:

PR #427, opened 2017-03-01
Remove lots of hardcoded values for config, and adds the `authenticate` hook

myknbani:
Fixes typedefs from #412.


myknbani:
Also, #412 defined legacy hooks, while there's only one hook in 1.x.

PR #433, opened 2017-03-05
Add client.js back for consistency

daffl:
Currently all client side modules are provided as `feathers-<name>/client` (e.g. `feathers/client`, `feathers-rest/client` etc.). Adding back a `feathers-authentication/client` which re-exports `feathers-authentication-client` in order to be consistent until we split everything into their own client modules (`feathers-rest-client`, `feathers-socketio-client` etc.) in the next version (see https://github.com/feathersjs/feathers-client/issues/137).

### DIFF
--- a/client.js
+++ b/client.js
@@ -1,0 +1,1 @@
+module.exports = require('feathers-authentication-client');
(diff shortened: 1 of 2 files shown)


ekryski:
:shipit: yup we should pull them all at once.

PR #442, opened 2017-03-10
Windows compatible version of the original compile comand with public folder support.

appurist:
- In the `legacy` branch, the `npm run compile` command was:
```
rm -rf lib/ && babel -d lib/ src/ && mkdir lib/public/ && cp src/public/* lib/public/
```
But that wasn't windows compatible. Two problems: forward slashes on the mkdir and cp commands cause parameter errors, and missing the `cp` command on Windows.  Although `rimraf` takes care of the first half, I suspect the `public` copy was lost because a cross-platform recursive mkdir+cp is hard to find. But it turns out that `babel -D` handles that.
- Also fixed the extra `-rf` that should not be used with `rimraf`.
- Also added `.idea` and `.vs` Windows IDE user folders to `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ _backup/
 
 # Yarn lockfile
 yarn.lock
+
+# local options/storage for JetBrains WebStorm and Visual Studio 2017 IDE folder
+.idea/
+.vs/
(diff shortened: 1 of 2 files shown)


daffl:
Thank you! I thought we got them all. Why do we have to copy a public folder? I can't see one in master.

PR #443, opened 2017-03-10
Windows tooling compatibility (rimraf, paths)

appurist:
- Windows compatible version of the original 'compile' comand with 'public' folder support:
  - replaced `rm -rf` with `rimraf`
  - replaced the `mkdir` and `cp` with `babel -D` which handles both, as well as paths with forward slashes (which breaks mkdir on Windows).

- Updated `.gitignores` to support Windows IDEs
  - .idea and .vs subfolders ignored now
  - also brought over the useful `.gitignore` updates from `master` for things like yarn.lock


appurist:
I don't know why Travis is complaining here. I've pulled https://github.com/appurist/feathers-authentication on a new Linux VPS and it compiles and tests without error. Same on Windows (after this PR).

I don't know enough about Travis but if you follow the [Details](https://travis-ci.org/feathersjs/feathers-authentication/builds/209638401) link above, it shows 1 of 3 passing (so that's probably closest to my setup on both platforms).  I'm not sure what those tasks are doing, and what is different between them.  Maybe different Linux environments?

appurist:
These were small, non-code changes, that test successfully on my local Windows and Linux test machines.  So I'm going to try this again more carefully, one change at a time. The .gitignores were just minor improvements to help Windows copies avoid showing unstaged changes after IDE loads, so optional. I'll take a look at that only after the main fix to the package.json file merges fine.  I'll do new separate PRs as each change is accepted by Travis.

PR #449, opened 2017-03-20
add option to allowUnauthenticated

MichaelErmer:
This adds to option `allowUnauthenticated` which makes it possible to authenticate the user, but also allow anonymouse users to pass the hook too.

```
auth.hooks.authenticate('jwt', {allowUnauthenticated: true})
```

Its required if you want to allow certain stuff to be done by anonymous users.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ _backup/
 
 # Yarn lockfile
 yarn.lock
+
+# local options/storage for JetBrains WebStorm and Visual Studio 2017 IDE folder
+.idea/
+.vs/
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@ node_modules
 # Users Environment Variables
 .lock-wscript
 
-lib/
 coverage/
 data.db
 _backup/
--- a/lib/express/authenticate.js
+++ b/lib/express/authenticate.js
@@ -1,0 +1,91 @@
+'use strict';
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+exports.default = authenticate;
+
+var _feathersErrors = require('feathers-errors');
+
+var _feathersErrors2 = _interopRequireDefault(_feathersErrors);
+
+var _debug = require('debug');
+
+var _debug2 = _interopRequireDefault(_debug);
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+var debug = (0, _debug2.default)('feathers-authentication:express:authenticate');
+
+function authenticate(strategy) {
+  var options = arguments.length > 1 && arguments[1] !== undefined ? arguments[1] : {};
+
+  // TODO (EK): Support arrays of strategies
+
+  if (!strategy) {
+    throw new Error('The \'authenticate\' hook requires one of your registered passport strategies.');
+  }
+
+  return function (req, res, next) {
+    // If we are already authenticated skip
+    if (req.authenticated) {
+      return next();
+    }
+
+    // if (!req.app.passport._strategy(strategy)) {
+    //   return next(new Error(`Your '${strategy}' authentication strategy is not registered with passport.`));
+    // }
+    // TODO (EK): Can we do something in here to get away
+    // from express-session for OAuth1?
+    // TODO (EK): Handle chaining multiple strategies
+    req.app.authenticate(strategy, options)(req).then(function () {
+      var result = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
+
+      // TODO (EK): Support passport failureFlash
+      // TODO (EK): Support passport successFlash
+      if (result.success) {
+        Object.assign(req, { authenticated: true }, result.data);
+        Object.assign(req.feathers, { authenticated: true }, result.data);
+
+        if (options.successRedirect && !options.__oauth) {
+          debug('Redirecting to ' + options.successRedirect);
+          res.status(302);
+          return res.redirect(options.successRedirect);
+        }
+
+        return next();
+      }
+
+      if (result.fail) {
+        if (options.failureRedirect && !options.__oauth) {
+          debug('Redirecting to ' + options.failureRedirect);
+          res.status(302);
+          return res.redirect(options.failureRedirect);
+        }
+
+        var challenge = result.challenge,
+            _result$status = result.status,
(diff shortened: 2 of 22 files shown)


MichaelErmer:
any comments?

marshallswain:
I've thought about this, recently.  (even before this PR)  We definitely want to allow this feature.  I was originally thinking of actually creating an anonymous auth strategy.  Actually, I think I'm still leaning in the direction of explicitly requiring `feathersClient.authenticate({strategy: 'anonymous'})`,.  It's basically the same feature as what you're enabling here, but with an API that more closely matches the API that devs are already using.

ekryski:
Thanks for the PR @MichaelErmer this didn't get missed I've been mulling it over for a while as well.

I don't see any reason why you can't just use the `iff` or `unless` common hook to achieve this, and have much more control.

```js
hooks.iff(
  hook => !!hook.data.accessToken, // check if a token is present (or whatever condition you want really)
  auth.hooks.authenticate('jwt')
)
```

Maybe I'm missing something but would that work @MichaelErmer?

MichaelErmer:
@ekryski Your solution works, however it's not login method agnostic as it uses specific values that might change in a future release, while the option can easily be keept compatible. I don't think it's good for such a standard features to require it to be wrapped every time it's used, it will  just bloat the code.

@marshall will it also authenticate the user if a token is present?

I think the 0.x approach, separating authenticate and the reject function was more convenient, is there a reason this was changed?

marshallswain:
Yeah, the setup would look like this:

```js
// This would be in a before authentication hook
auth.hooks.authenticate(['anonymous', 'jwt']);
```

Then you call out to the anonymous strategy with
```js
feathersClient.authenticate({strategy: 'anonymous'});
```

Then you'd get back an anonymous JWT.  We'd probably want to put some type of identifier in it for keeping track of anonymous users across sessions.  Something like `{anonymousId: 'some-id'}

Then on all other services, you could just use the `jwt` strategy, and in hooks you'd handle tokens from anonymous users differently than tokens from authenticated users.

MichaelErmer:
@marshallswain but that would require the client to "authenticate", we have some services that are read only and should be available to all users, we don't want to tell them to use feathers as client as its up to them.

My solution allows the service to be accessed without bothering about authentication at all, we have hooks that filter the returned fields based on params.authenticated.

ekryski:
@marshallswain yeah I think you were missing @MichaelErmer's point there. He wants user's to be able to access a service regardless of whether they are authenticated or not. I'm just guessing that when they are authenticated they get to see specific data.

@marshallswain for "anonymous" auth like you were mentioning you don't need to do anything. No custom strategy. You can already do that by simply removing the `auth.hooks.authenticate` hook from before the `/authentication` service and you'll get back an "anonymous" JWT access token which you could use to access services.

---

> Your solution works, however it's not login method agnostic as it uses specific values that might change in a future release, while the option can easily be kept compatible.

@MichaelErmer which "specific values" are you referring to? There aren't any plans to change `accessToken`. The beauty of the method I provided is that it allows you to easily customize which conditions are used to determine whether auth should run.

We're starting to move away from hooks that have conditions for execution buried inside them. It's hard for people to customize and when you get a lot of them makes it hard to determine the **actual** code execution that is happening. When we use hooks like `iff` and `unless` to bring conditional execution outside of the hook function it:

1. Keeps the hook cleaner
2. Reduces cognitive overhead
3. Allows for more flexible combinations of hooks to create more complex "workflow"
4. Makes it easier to see the code flow by just looking at your main `hooks/index.js` file

All that being said, could you elaborate on your on your use case? I think I might be seeing what you are looking for. Essentially I think you want it to not error if authentication fails. If that's the case we'd need to make some changes to the PR because you want it to "pass through" not just set `hook.params.authenticated = true`.

marshallswain:
Yep, I definitely misunderstood.

@ekryski But there's no way to do both anonymous auth (Just one step away from unauthenticated) and any other type of auth.  But either way, this issue has nothing to do with it.

MichaelErmer:
@ekryski Yes, it should pass through and not set authenticated if the token is not present/invalid. 
If you want to keep hooks cleaner, wouldnt it be better if the redirect and throw stuff are separate hooks? this would also eliminate my PR.
Please check the last version of the code it sets hook.params.authenticated correctly.

ekryski:
@MichaelErmer did you push your latest updates? Looks like it's pretty much the same?? Did I miss something?

MichaelErmer:
@ekryski what else would you like to see in the PR? it only needs the 3 changed lines to work.

MichaelErmer:
Any news on this?

MichaelErmer:
ping

ekryski:
Hey @MichaelErmer we're going to be addressing this in a different way in the next version of Auth right around the corner when we make the authentication service and actual service and make it more transport agnostic. The JWT creation will happen in hooks so you'll have total control over when and what type of token is created instead of having to do a monkey patch like this.

We're also in the process up updating all the core repos so that they aren't transpiled and you can fork and install them from Github. All of this is happening in the very near term so I'm going to close this PR.

Thank you so so so much for taking the time and the thought. You can follow progress of the auth update [here](https://github.com/feathersjs/feathers-authentication/projects) and we'd love your input in Slack in the beta testers channel. I'll be updating that Github project over the next week or two to outline the roadmap. 😄  🍻

subodhpareek18:
I have made a custom passport strategy for anonymous users, which gives them a default role and creates a user before returning a token.

```js
// anonymous.js

import Strategy from 'passport-custom';

export default opts => {
  return function() {
    const verifier = async (req, done) => {
      const role = await this.service(opts.roleService)
        .find({
          query: {
            name: opts.defaultRole,
            $limit: 1
          }
        })
        .then(res => res.data[0]);

      const user = await this.service(opts.userService).create({
        roleId: role.id
      });

      return done(null, user, {
        userId: user.id
      });
    };

    this.passport.use('anonymous', new Strategy(verifier));
  };
};
```
```js
    this.configure(auth(server.auth));
    this.configure(jwt());
    this.configure(local(server.auth.local));
    this.configure(anonymous(server.auth.anonymous));
```

```
POST /authentication
{
  strategy: 'anonymous'
}
```
I haven't yet, but will conditionally only allow this type of requests from a specific customer facing server.

----

I had another example of a apiKey based strategy, will share gists for both of them or add a PR to the docs so anyone can create custom strategies easily and the standard passport way.

PR #450, opened 2017-03-21
Add a 'is owner or restrict changes hook'

codingfriend1:
Only the owner of the document has the ability to modify their document without restriction. Unauthenticated users or users that don't own the document can still make changes but are restricted to using certain properties.

### DIFF
--- a/src/hooks/index.js
+++ b/src/hooks/index.js
@@ -9,6 +9,7 @@ import verifyToken from './verify-token';
 import verifyOrRestrict from './verify-or-restrict';
 import populateOrRestrict from './populate-or-restrict';
 import hasRoleOrRestrict from './has-role-or-restrict';
+import ownerOrRestrictChanges from './owner-or-restrict-changes';
 
 let hooks = {
   associateCurrentUser,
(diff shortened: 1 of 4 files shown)


ekryski:
@codingfriend1 awesome PR! Thanks for putting in so much work here!

We're not going to be bringing additional functionality into the legacy branch. We're wrapping up auth docs and are encouraging people to move towards `feathers-authentication@1.x` and `feathers-permission`.

However, this could totally be merged into [feathers-legacy-authentication-hooks](https://github.com/feathersjs/feathers-legacy-authentication-hooks) (which can also be used with auth 1.x).

I like what you did by limiting certain properties. I'd like to see if we can incorporate that into feathers-permissions as well.

If you are up for creating a PR to the legacy-hooks repo I'll merge it quickly.

Thanks for taking the time to put this together! 😄

daffl:
I still feel like `feathers-permissions` would be overkill if you just want to restrict the authenticated user to editing, removing and listing their records. I ended up writing a very similar hook to this one myself, too.

ekryski:
Closed as this was merged here: https://github.com/feathersjs/feathers-authentication-hooks/pull/2

Issue #448, opened 2017-03-18
Add ability to control if socket is marked as authenticated.


marshallswain:
I'm making a plugin that enables the workflow shown in the diagram, below.  I'm overloading the `authenticate` method with strategies built on `feathers-authentication-local` with custom Verifiers, then using hooks for cleanup work.

```js
app.authenticate({strategy: 'email', email, signature}) // returns a salt & challenge
app.authenticate({strategy: 'challenge', challenge, signature}) // returns an accessToken & secret
app.authenticate({strategy: 'two-factor', type}) // Requests that a 2FA code be sent.
app.authenticate({strategy: 'two-factor', signature}) // Authenticate with hash based on 2FA
```

What I need is the ability to conditionally mark the socket as authenticated.  I think adding an `{authenticated: false}` to the params in a before hook would be good enough.

```js
app.service('authentication').hooks({
  before: {
    create: [
      auth.hooks.authenticate(['email', 'challenge', 'two-factor', 'jwt']),
      function (hook) {
        if (hook.data.strategy === 'email') {
          hook.params.authenticated = false
        }
      }
    ]
  }
})
```

Then we add a condition to the socket handler to check for this before marking the socket as authenticated, here: https://github.com/feathersjs/feathers-authentication/blob/master/src/socket/handler.js#L128

This basically opens the door for multi-factor auth solutions to be built: https://github.com/feathersjs/feathers-authentication/issues/130

Some of the facts may not be quite straight in this diagram, but the basic idea is that the user can authenticate without having to ever send his/her password.

![screen shot 2017-03-18 at 8 15 30 am](https://cloud.githubusercontent.com/assets/128857/24072883/6367ff7e-0bb3-11e7-94d1-02bd9b1fd319.jpg)

ekryski:
Going to put what I sent to you in Slack here as well....

I follow what you've got here, but I'm wondering if re-using the `auth-local` strategy for all of those steps is really the right way to do this. It seems like it really should just be a custom passport strategy to handle all that hand shaking and then when all of it is done, it resolves so that the strategy calls back with a success.

I've done both multi-factor auth and "magic-link" auth in the same fashion you have (by overriding `feathers-authentication-local`). It works, but it's kind of hacky. In my mind they really should be passport strategies that you register.

My concern (a small one) with adding that bypass is that someone inadvertently sets that param and it causes weird side effects. I want to move away from having readable/writeable params on `hook.params` that we are using internally. They need to be more protected.

[...]

Issue #432, opened 2017-03-05
`hook.params.provider` is not set when calling `client.authenticate()`


huafu:
When calling `client.authenticate()` in the client (here I'm using `socket.io`), the `authenticate.create` service method is called on the server, but `hook.params.provider` is `undefined` while I think it should be set, even tho the `client.authenticate()` is triggering this service call thru a custom socket event. Same for `client.logout()`.

I am using `1.0.2` version of `feathers-authentication`.

ekryski:
Agreed. Thanks for the issue. I'll try and tackle that one quickly.

marshallswain:
Fixed in #459

daffl:
Going to close this we'll release shortly.

IamCarbonMan:
Any update on this? It's been several months and this is preventing me from being able to do

    commonHooks.when(hook => hooks.params.provider, commonHooks.discard("password"))

which is pretty necessary for proper security.

daffl:
This has been released in v1.2.0 (see https://github.com/feathersjs/feathers-authentication/blob/master/CHANGELOG.md#v120-2017-03-23)

IamCarbonMan:
I'm using 1.2.7 and still having the same issue.

daffl:
@marshallswain Any thoughts?

IamCarbonMan:
My setup is definitely not the standard feathers-cli generated app, so it might be something else. The server is a single file, here it is in a [gist](https://gist.github.com/IamCarbonMan/f1374ecb0fb6d1adb8fd57158d55fad5). The actual error I get in the browser when calling client.authenticate is "cannot read property provider of undefined", suggesting that hook.params is itself undefined for some reason.

daffl:
You have a typo in https://gist.github.com/IamCarbonMan/f1374ecb0fb6d1adb8fd57158d55fad5#file-server-js-L16. It should be `hook.params.provider` not `hooks.params.provider` (hence the error you were getting).

PR #459, opened 2017-03-22
Fix socket auth

### LINKED ISSUE #455: 1.0 authentication service hooks don't run when client uses feathers-socketio
The authentication service hooks don't get hit when I use the `feathers-socketio` transport.  In the example, below, the hook only runs if I use `feathers-rest`.

```js
  before: {
    create: [function (hook) {
      // OOPS, this only runs during requests from feathers-rest
      console.log(hook);
    }]
  },
  after: {
    create: [authHook(options)]
  }
});
```

marshallswain:
Closes #455, #448, and #432

I left a couple of `debugger`s in that aren't getting it in the rest.test.js file when testing events  (search for `it.only`).

This change channels socket auth through the token service instead of calling app.authenticate manually before hitting the token service.

This also makes it so the client has to provide a strategy explicitly, and removes the strategy iteration from `passport/authenticate.js`.

Also updates tests.

### DIFF
--- a/src/express/authenticate.js
+++ b/src/express/authenticate.js
@@ -11,7 +11,6 @@ export default function authenticate (strategy, options = {}) {
 
   return function (req, res, next) {
     // If we are already authenticated skip
-    debugger;
     if (req.authenticated) {
       return next();
     }
@@ -23,7 +22,6 @@ export default function authenticate (strategy, options = {}) {
     // from express-session for OAuth1?
     // TODO (EK): Handle chaining multiple strategies
     req.app.authenticate(strategy, options)(req).then((result = {}) => {
-
       // TODO (EK): Support passport failureFlash
       // TODO (EK): Support passport successFlash
       if (result.success) {
--- a/src/express/authenticate.js
+++ b/src/express/authenticate.js
@@ -11,6 +11,7 @@ export default function authenticate (strategy, options = {}) {
 
   return function (req, res, next) {
     // If we are already authenticated skip
+    debugger;
     if (req.authenticated) {
       return next();
     }
@@ -22,6 +23,7 @@ export default function authenticate (strategy, options = {}) {
     // from express-session for OAuth1?
     // TODO (EK): Handle chaining multiple strategies
     req.app.authenticate(strategy, options)(req).then((result = {}) => {
+
       // TODO (EK): Support passport failureFlash
       // TODO (EK): Support passport successFlash
       if (result.success) {
--- a/src/hooks/authenticate.js
+++ b/src/hooks/authenticate.js
@@ -1,9 +1,10 @@
 import errors from 'feathers-errors';
 import Debug from 'debug';
+import merge from 'lodash.merge';
 const debug = Debug('feathers-authentication:hooks:authenticate');
 
-export default function authenticate (strategy, options = {}) {
-  if (!strategy) {
+export default function authenticate (strategies, options = {}) {
+  if (!strategies) {
     throw new Error(`The 'authenticate' hook requires one of your registered passport strategies.`);
   }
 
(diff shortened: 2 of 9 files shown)


src/socket/handler.js
```diff
@@ -57,135 +55,94 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
...
       const { strategy } = data;
-      const body = omit(data, 'strategy');
-
+      // const body = omit(data, 'strategy');
       socket._feathers = {
         query: {},
-        params: {},
-        body,
+        provider: 'socketio',
+        // body: data,
```
marshallswain: Needs to be removed?

test/integration/rest.test.js
```diff
@@ -441,7 +441,7 @@ describe('REST authentication', function () {
     });
 
     describe('when authentication succeeds', () => {
-      it('emits login event', done => {
+      it.only('emits login event', done => {
```
marshallswain: This is the failing test.  Well, one of two that are failing for what looks like the same cause.

src/express/authenticate.js
```diff
@@ -11,6 +11,7 @@ export default function authenticate (strategy, options = {}) {
 
   return function (req, res, next) {
     // If we are already authenticated skip
+    debugger;
```
marshallswain: This `debugger` never gets hit.

src/hooks/authenticate.js
```diff
@@ -19,13 +20,29 @@ export default function authenticate (strategy, options = {}) {
...
+        strategy = strategies;
+      }
+    }
+
+    // Handle the case where authenticate hook was registered without a passport strategy specified
+    if (!strategy) {
+      return Promise.reject(new errors.GeneralError(`You must provide an authentication 'strategy'`));
+    }
+
     // NOTE (EK): Bring this in when we decide to make the strategy required by the client
```
ekryski: We can remove this note. `strategy` is required now.

src/passport/authenticate.js
```diff
@@ -21,10 +21,21 @@ export default function authenticate (options = {}) {
         // Allow you to set a location for the success payload.
         // Default is hook.params.user, req.user and socket.user.
         const entity = strategyOptions.entity || strategyOptions.assignProperty || options.entity;
-        let failures = [];
-        let strategies;
+        let strategyName = request.body && request.body.strategy;
```
ekryski: This should probably be:

```js
request.body = request.body || {};
let stategyName = request.body.strategy;
```

src/socket/handler.js
```diff
@@ -57,135 +55,92 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
...
-      if (!strategy) {
-        const error = new errors.BadRequest(`An authentication 'strategy' must be provided.`);
-        return callback(normalizeError(error));
-      }
-
-      if (!app.passport._strategy(strategy)) {
-        const error = new Error(`Your '${strategy}' authentication strategy is not registered with passport.`);
-        return callback(normalizeError(error));
-      }
+      // if (!strategy) {
```
ekryski: This can be removed.

src/socket/handler.js
```diff
@@ -57,135 +55,92 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
...
-            //   return {
-            //     redirect: true,
-            //     status: 302,
-            //     url: stategyOptions.successRedirect
-            //   };
-            // }
-            return Promise.resolve(result.data);
-          }
+      // if (!app.passport._strategy(strategy)) {
+      //   const error = new Error(`Your '${strategy}' authentication strategy is not registered with passport.`);
```
ekryski: This can also be removed

src/socket/handler.js
```diff
@@ -57,135 +55,92 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
...
-          // }
-
-          // NOTE (EK): This handles redirects and .pass()
-          return Promise.reject(new errors.NotAuthenticated('Authentication could not complete. You might be using an unsupported socket authentication strategy. Refer to docs.feathersjs.com for more details.'));
-        })
-        .then(result => {
-          // Now that we are authenticated create our JWT access token
-          const params = Object.assign({ authenticated: true }, result);
-          return service.create(result, params).then(tokens => {
+      // const promise = app.authenticate(strategy, strategyOptions)(socket._feathers)
```
ekryski: This can be removed

PR #461, opened 2017-03-23
Filter out all events

### LINKED ISSUE #460: Add event filtering by default (socket.io)
"authentication created" events are propagated to everyone (with the accessToken), even with logged out users. I think this should be filtered/silenced by default.

daffl:
Closes #460

### DIFF
--- a/src/service.js
+++ b/src/service.js
@@ -59,6 +59,12 @@ export default function init (options) {
       successRedirect(),
       failureRedirect(options)
     );
+
+    const service = app.service(path);
+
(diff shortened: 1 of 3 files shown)


marshallswain:
Looks good.  Do you think we need to pad the timing a little bit to make sure there has been enough time for an event to occur before it calls `done()`?

daffl:
Mocha will just complain that `done` was called twice without this fix.

ekryski:
:shipit:

PR #464, opened 2017-03-25
Fix Oauth callback receive array instead of model instance

hoangzinh:
Some of service adapters like feathers-sequelize or feathers-moongo implemented `patch` action that will return an array of instances.

For example feathers-sequelize, this is the way we're using to updating database in `patch` action

[https://github.com/feathersjs/feathers-sequelize/blob/master/src/index.js#L134](https://github.com/feathersjs/feathers-sequelize/blob/master/src/index.js#L134)

Follow documentation of sequelize, 
[http://docs.sequelizejs.com/en/latest/api/model/#updatevalues-options-promisearrayaffectedcount-affectedrows](http://docs.sequelizejs.com/en/latest/api/model/#updatevalues-options-promisearrayaffectedcount-affectedrows)

The `update` method of Model will update multiple instances that match the where options and returns an array of those instances.

I knew that we're moving to `v.1.0` but this's a bug in legacy repo

### DIFF
--- a/src/services/oauth2.js
+++ b/src/services/oauth2.js
@@ -61,7 +61,7 @@ export class Service {
           debug(`Updating user: ${id}`);
 
           return app.service(options.userEndpoint).patch(id, data).then(updatedUser => {
-            return done(null, updatedUser);
+            return done(null, updatedUser[0]);
           }).catch(done);
         }
 


daffl:
This might be a bug with Sequelize or the `id` is `null` which is also not correct - did you set the id property accordingly? A `patch` with an id should never return an array.

hoangzinh:
@daffl , I sure that I setup the `id` correct. The problem as I said in the above link,
```
      return this.Model.update(omit(data, this.id), options)
            .then(results => results[1])
            .then(select(params, this.id))
            .catch(utils.errorHandler);
```
We used method `update` in Sequelize Model and as it's described in their document. 

```
Update multiple instances that match the where options. The promise returns an array with one or two elements. The first element is always the number of affected rows, while the second element is the actual affected rows
```

ekryski:
@hoangzinh what is the value of `id` when you are trying to update the user? Is it `null`?

hoangzinh:
Okay, I will provide more details how did I produce that issue.
Versions:
1. Node: v6.3.1
2. Feathersjs: 2.1.1
3. feathers-rest: 1.7.1
4. feathers-sequelize: 1.4.3
5. feathers-authentication: 0.7.12

I use feathers-cli to generate a service called User, then make a `PATCH` request with endpoint
`/users/1` and get an array: [user1].

I used built-in rest service && I really make sure that the user with id =1 is exists so I think the value of `id` should be not null.
It's also weird with me in the first place as `PUT` return the single resource but `PATCH` return an array.

p/s: I really not sure is it a bug on featherjs-sequelize OR feathers-authentication is out of date with featherjs-sequelize.

ekryski:
We haven't seen this issue come up again and since this is breaking tests I'm going to close this PR. Thanks for contributing! 😄

PR #474, opened 2017-04-06
Allow the cookie to be set if action is not `remove`

marshallswain:
This removes the service requirement for setting the cookie.  If it did run after `remove` it won’t set the cookie, though.

Closes https://github.com/feathersjs/feathers-authentication-oauth2/issues/12

### DIFF
--- a/src/express/set-cookie.js
+++ b/src/express/set-cookie.js
@@ -22,9 +22,12 @@ export default function setCookie (authOptions = {}) {
     if (options.enabled && options.name) {
       const cookie = options.name;
 
-      // Only set the cookie if this was called after a service method and
-      // we weren't removing the token and we have a JWT access token.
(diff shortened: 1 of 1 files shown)


src/express/set-cookie.js
```diff
@@ -22,9 +22,12 @@ export default function setCookie (authOptions = {}) {
     if (options.enabled && options.name) {
       const cookie = options.name;
 
-      // Only set the cookie if this was called after a service method and
-      // we weren't removing the token and we have a JWT access token.
-      if (res.hook && res.hook.method !== 'remove' && res.data && res.data.accessToken) {
+      // Don't set the cookie if this was called after removing the token.
+      if (res.hook && res.hook.method === 'remove') {
```
ekryski: Yup I think this should be fine. Good catch on that issue. Funny how simple logic can have such implications.

ekryski:
:shipit:

Issue #458, opened 2017-03-22
Overriding JWT's expiresIn with a value more than 20d prevents users from signing in


johncrisostomo:
**UPDATE:** I forked this and tried running the tests locally, inspecting the tokens generated, and they seem to have a correct `exp` value. What gives?

I am trying to make a token expire a little longer than the default '1d', however, when I try a value greater than '20d', users are no longer able to sign in the app. I noticed that on the `utils.js` test cases, a value of '1y' was used but the token created was never 'asserted' to have a valid `exp` value.

I am aware that having a long enough token expiry is bad for security reasons, but I am just curious why this is not working properly.

johncrisostomo:
I re-opened this issue because I really want to know what's going on. Thank you in advance.

ekryski:
Hey @johncrisostomo,

I honestly have no idea. It **may** be something to do with the underlying [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) library that is being used. We're not doing anything to limit that and there isn't any issue or mention of a cap on the TTL of a JWT in the jsonwebtoken lib.

As you said, we don't recommend having tokens that long anyway but you might want to ask over on the [node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) module if there is a limitation on the `expiresIn` field.

jamesholcomb:
@ekryski Are you recommending refresh token strategy in lieu of a long expiration token?

marshallswain:
@jamesholcomb there's no blanket recommendation, really.  It depends on each service's security requirements. Generally, using refresh tokens on a service will allow you to take advantage of JWT auth not requiring a round trip to the db.  Shorter expirations and using refresh tokens will probably be more secure than long expirations.

jamesholcomb:
I need to secure every service endpoint with a JWT and short expiration (1d).   Do I have to roll my own refresh token solution?

marshallswain:
@jamesholcomb You can either roll your own, or if you're on a low-traffic app, query the db in a hook and check that the user hasn't been blocked.  We will eventually release an official solution.

johncrisostomo:
@ekryski That's the reason why I also got confused. I checked the tokens generated under "when adding custom options" in `utils.test.js` and it's giving me a valid token. So I kind of know that the issue isn't on feathers-authentication alone, but with somewhere else. I'll look into it more this weekend as I am really curious as to what is causing it.

asdacap:
[...]

PR #499, opened 2017-05-08
Fix issue with very large token timeout.

asdacap:
### Summary

Fix the socket authentication immediately logout when setting a very large token expiry time. See #458. This is likely due to 32bit integer overflow in `setTimeout`. It overcome this issue by using the `long-timout` npm package.


daffl:
This looks good to me, thank you! Once @ekryski or @marshallswain give the ok we'll release it.

marshallswain:
Please :shipit:

ekryski:
This looks great! Thanks @asdacap! ❤️

PR #506, opened 2017-05-10
Fix default authentication config keys

ekryski:
### Summary
The default config key was changed to `authentication` and was not updated here. This fixes that. This isn't a breaking change for anyone using v2 of the generator and/or are passing their auth config explicitly when configuring auth plugins:

```js
const config = app.get('auth');
```

but it will be for people that migrated to the new version of auth and still have a config file with:

```js
"auth": {
   // auth config values
}
```

instead of:

```js
"authentication": {
   // auth config values
}
```

For that reason, this and the associated PRs will need to be major releases:

- https://github.com/feathersjs/feathers-authentication-jwt/pull/21
- https://github.com/feathersjs/feathers-authentication-local/pull/22
- https://github.com/feathersjs/feathers-authentication-oauth1/pull/18
- https://github.com/feathersjs/feathers-authentication-oauth2/pull/24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 # Change Log
 
-## [Unreleased](https://github.com/feathersjs/feathers-authentication/tree/HEAD)
-
-[Full Changelog](https://github.com/feathersjs/feathers-authentication/compare/v1.2.2...HEAD)
+## [v1.2.3](https://github.com/feathersjs/feathers-authentication/tree/v1.2.3) (2017-05-10)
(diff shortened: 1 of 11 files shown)


marshallswain:
Why not ease the burden with a `var authConfig = app.get('auth') || app.get('authentication')` and merge that in, instead?  We'll have some cleanup work to do for tech support, otherwise.

ekryski:
@marshallswain Ah shit sorry. I didn't even see your comment. That's a good idea.

ekryski:
:shipit:

Issue #418, opened 2017-02-11
Prefixing socket events


devel-pa:
Hi,
Trying to integrate some puzzle pieces from different sets I found an issue, not necessarily a feathers only issue, but I think it would be a nice to have: prefixing socket events to avoid event collisions

https://github.com/feathersjs/feathers-authentication/blob/master/src/socket/handler.js#L194

Configurable or not, to have instead of `socket.on('authenticate', authenticate);` something like `socket.on('fjs-authenticate', authenticate);`

Best,
Paul

daffl:
`authenticate` is a pretty common event to use for Websocket authentication and the only special event outside of its services sent by Feathers. I'm not sure it is worth adding another configuration option since you probably won't have two different authentication mechanisms in your app.

daffl:
I'm going to close this since we probably won't take this on. We'll definitely accept or help out with a pull request adding this option though.

devel-pa:
Sorry for the delay, I have some health issues.
My problem with naming events is that sometimes 2 libraries can have same event, and even if it's not really used it have side effects.
As ex:
http://socketcluster.io/#!/docs/api-scsocket-client
where `authenticate` event does not pass to server as it's considered an internal event. Not a feathers issue, but more namespaced events (as long as they are transparent) I think are better.

As soon as I can, I'll submit a PR.

ekryski:
We're going to be converting the authentication service to a proper Service which will make it so that the event is `<service_name> created`. Which should nullify this issue.

PR #507, opened 2017-05-10
Prefixing socket events.

devel-pa:
https://github.com/feathersjs/feathers-authentication/issues/418

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -5,6 +5,7 @@ const defaults = {
   header: 'Authorization',
   entity: 'user',
   service: 'users',
+  socketEventPrefix: '',
   passReqToCallback: true,
   session: false,
   cookie: {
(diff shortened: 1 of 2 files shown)


ekryski:
Thanks for the PR @devel-pa! 😄  This needs some tests before we can merge it.

I also wanted to pose a question here. Does it make sense to allow people to add their own prefix or should we just allow them to call the event whatever they want?

devel-pa:
Hi @ekryski,
The reason I need it is that sometimes I have conflicts with other libraries that have the same `authenticate` event. 
As it is hardcoded everywhere, I thought that is better to prefix it as it remains meaningful and simple, without changing much. Just a small flexibility that avoids sometimes naming issues.
Best,
Paul

ekryski:
Going to close this as per my comment here: https://github.com/feathersjs/feathers-authentication/issues/418#issuecomment-317828258.

Regardless, thanks for taking the time to submit the PR! ❤️

Issue #293, opened 2016-09-17
User (Entity) needs to be updated on the socket after authentication


sunabozu:
So here is the problem. 
1. I authenticate
2. I have a required role to create a particular resource, protected by the hook, and I create it successfully
3. I edit my database (both directly and through the client), removing the required role from the user document
4. I try to create the same resource and I succeed (<== problem)
5. I logout, authenticate again and now the hook is working, I can't create the resource

So it seems like it caches the user object from the database or (which is much worse) just checks the object sent by the user himself. In the later case it's a security flaw, since any user can assign himself any role on the client side.

marshallswain:
@sunabozo, are you modifying the token payload at all?  As in, are you putting storing permissions in the token?  Where is the auth getting its stale data?

sunabozu:
@marshallswain I don't touch the token. I use `restrictToRoles` in the list of the `before` hooks for my model. Particularly for the `create` method:

``` js
exports.before = {
  all: [],
  find: [],
  get: [],
  create: [
    authHooks.restrictToRoles({
      roles: ['admin']
    })
  ]
};
```
The `user` model has a `roles` field, where the roles are stored. So I modify that field for the current user and nothing else.
I hope I understood your questions correctly.

marshallswain:
Yeah, looks like you understood what I was trying to say.  What hooks are you running before the `restrictToRoles` hook?

marshallswain:
I recommend using Visual Studio Code's debugger and setting a breakpoint here: https://github.com/feathersjs/feathers-authentication/blob/master/src/hooks/restrict-to-roles.js#L32  To see if you're still getting stale data, somehow.  Using the populateUser hook just before this one should query the database for fresh data every time, unless you have some kind of cacheing mechanism in place.  Not sure what your db setup is.

marshallswain:
Setting up VSCode is a one-line change with a generated app.  Just press the green play button to start the debugger for the first time and it will prompt you for a configuration type.  Choose node, then change the highlighted line to look like this:

![screen shot 2016-09-17 at 1 00 09 pm](https://cloud.githubusercontent.com/assets/128857/18610334/ce885298-7cd6-11e6-9a3e-61905ce37e80.jpg)

Then you can just open that file in your node_modules and set a break point on that line.  Run the query and it will stop at that point so you can inspect just like with Chrome debug tools.

sunabozu:
[...]

PR #521, opened 2017-06-01
Update socket entity

marshallswain:
Fixes: https://github.com/feathersjs/feathers-authentication/issues/293

This updates the socket entity for either primus or socket.io.  (it would need a small refactor to work for both)

### DIFF
--- a/src/socket/handler.js
+++ b/src/socket/handler.js
@@ -2,6 +2,7 @@ import Debug from 'debug';
 import ms from 'ms';
 import { normalizeError } from 'feathers-socket-commons/lib/utils';
 import lt from 'long-timeout';
+import updateEntity from './update-entity';
 
 const debug = Debug('feathers-authentication:sockets:handler');
 
@@ -20,6 +21,7 @@ function handleSocketCallback (promise, callback) {
 export default function setupSocketHandler (app, options, { feathersParams, provider, emit, disconnect }) {
   const authSettings = app.get('auth');
   const service = app.service(authSettings.path);
+  const entityService = app.service(authSettings.service);
 
   return function (socket) {
     let logoutTimer;
@@ -139,5 +141,8 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
     socket.on('authenticate', authenticate);
     socket.on(disconnect, logout);
     socket.on('logout', logout);
+
+    entityService.on('updated', updateEntity);
+    entityService.on('patched', updateEntity);
   };
 }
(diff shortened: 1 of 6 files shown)


src/socket/update-entity.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+module.exports = function updateEntity (entity, meta) {
+  const { app } = meta;
+  const authConfig = app.get('auth');
+  var idField = app.service(authConfig.service).id;
```
ekryski: super knit pick so not a big deal. Use `let` instead of `var`? Same on line 14.

src/socket/handler.js
```diff
@@ -139,5 +141,8 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
     socket.on('authenticate', authenticate);
     socket.on(disconnect, logout);
     socket.on('logout', logout);
+
+    entityService.on('updated', updateEntity);
```
ekryski: The only thing I'm wondering about is whether we should do

```js
app.io.on(`${entity} updated`, updateEntity)
app.io.on(`${entity} patched`, updateEntity)
```

so that filters don't get applied. Since this is an internal listener we probably don't want any filters that the developer may have registered to apply here. Otherwise they might be inadvertently preventing an update.

However, there is also a trade-off with that. If we don't use the `app.service()` method then we can't support listening to any remote-decoupled services. In practice that might not be all that realistic and my idea of being able to make auth a standalone service separate from the entity might not actually be possible. Usually a `users` service and `auth` service live on the same app instance.

marshallswain:
Since filters don't run on the server, but operate at the provider level, I'm going to go ahead and merge this in.

jerfowler:
Noticed the comparison doesn't work when the entityId is an ObjectID (mongodb/mongoose) :

```JavaScript
if (entityId === socketEntityId) {
    Object.assign(socketEntity, entity);
}
```
Plus, using `Object.assign()` over the top of `socketEntity `doesn't remove properties if they were deleted from `entity`.

Also, getting errors when there are open unauthenticated connections, which I opened a new ticket here: #529 

Wondering if it would be better to do so something like this:

```JavaScript
Object.keys(socketMap).forEach(socketId => {
    const socket = socketMap[socketId];
    const feathers = socket.feathers || socket.request.feathers
    const socketEntity = feathers && feathers[authConfig.entity];

    if (socketEntity) {
      const socketEntityId = socketEntity[idField];

      if (`${entityId}` === `${socketEntityId}`) {
        feathers[authConfig.entity] = entity;
      }
    }
  });
```

jerfowler:
I created a pull request with some test cases, here: #531 

Similar to above with the exception I had to keep the `Object.assign()`. Looks like you need to assign to the existing entity object otherwise it breaks external references. So now I just delete the properties.

PR #522, opened 2017-06-01
Remove hook.data.payload

marshallswain:
This removes `hook.data.payload` because it’s too easy to open a security hole when implementing custom authentication solutions.  `hook.params.payload` is the only default mechanism to customize the JWT, now.  A hook can be used to restore this functionality, if needed.

### DIFF
--- a/docs/migrating.md
+++ b/docs/migrating.md
@@ -229,7 +229,7 @@ By default the payload for your JWT is simply your entity id (ie. `{ userId }`).
 function customizeJWTPayload() {
   return function(hook) {
     console.log('Customizing JWT Payload');
-    hook.data.payload = {
+    hook.params.payload = {
       // You need to make sure you have the right id.
       // You can put whatever you want to be encoded in
       // the JWT access token.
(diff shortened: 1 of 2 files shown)


ekryski:
Agreed. I think this was an oversight on my part and was a remnant left over from before I figured out how we should be returning the payload from the auth plugins. See [this line](https://github.com/feathersjs/feathers-authentication-local/blob/master/src/verifier.js#L79) as an example.

This may be a breaking change for people. I'm not sure if anyone has been using that but my guess is this probably should be a major release. Thoughts?

This is totally a :shipit: though.

marshallswain:
This might be a breaking change for one or two people.  In order to hit the vulnerability you had to use a custom verifier and not pass any payload out of the verifier.  Otherwise, the params.payload always overwrite the data.payload, anyway.  I think we should get it into the current release.

ekryski:
Yeah let’s ship it. Technically it should be a major since we are removing functionality. But that functionality shouldn’t be there and unless you were really monkeying with stuff in a way you shouldn’t be, that PR shouldn’t have an effect.

:shipit:

PR #585, opened 2017-10-13
Features/typescript fix

TimMensch:
### Summary

TypeScript 2.5.2 in strict mode throws many errors on the built-in `index.d.ts` declarations. It doesn't seem possible to prevent TypeScript from throwing these errors (the `exclude` option doesn't work). I feel it's bad form to "ignore" errors, so I'd rather seem them patched in the upstream source.

This is the second of three patches that fixes these errors. I will be submitting a pull request for feathers-hooks shortly.


ekryski:
Thanks!

PR #591, opened 2017-10-24
Update to new plugin infrastructure

daffl:
For https://github.com/feathersjs/feathers/issues/608


.istanbul.yml
```diff
@@ -11,7 +9,7 @@ reporting:
     - text
     - lcov
   watermarks:
-    statements: [70, 90]
-    lines: [70, 90]
-    functions: [70, 90]
-    branches: [70, 90]
\ No newline at end of file
+    statements: [50, 80]
```
ekryski: Don't want to bump these watermarks up? I think we should aim for > 90% or >95% coverage on core modules.

.npmignore
```diff
@@ -1,12 +1,10 @@
 .editorconfig
 .jshintrc
```
ekryski: probably need `.eslintrc.yml` in there instead of `.jshintrc`

index.d.ts
```diff
@@ -1,107 +0,0 @@
-// TODO:  don't have the defs from the hooks PR yet, fix later
```
ekryski: Is this moved to the https://github.com/feathersjs-ecosystem/feathers-typescript repo? Just curious.

package.json
```diff
@@ -27,26 +23,22 @@
...
     "release:major": "npm version major && npm publish",
-    "release:pre": "npm publish --tag next",
+    "release:pre": "npm publish --tag pre",
     "changelog": "github_changelog_generator && git add CHANGELOG.md && git commit -am \"Updating changelog\"",
-    "compile": "rimraf lib/ && babel -d lib/ src/",
-    "watch": "babel --watch -d lib/ src/",
     "lint": "semistandard --fix",
     "mocha": "mocha --opts mocha.opts",
-    "test": "npm run compile && npm run lint && npm run coverage && nsp check",
+    "test": "npm run lint && npm run coverage",
```
ekryski: Are we dropping the `nsp` check? Does Greenkeeper cover us there?

index.d.ts
daffl: Yeah, will be. I don't want to mess with our repos because TypeScript definitions are broken all the time. They'll be published separately and then pulled in.

.istanbul.yml
daffl: We can do that in bulk for all repos once they are done.

package.json
daffl: Nah, good call, added it back.

PR #599, opened 2017-11-02
add option to allowUnauthenticated

MichaelErmer:
Reopening of https://github.com/feathersjs/authentication/pull/449, without the transpiled version.

We really need this feature in the current version as we can not rewrite the entire project again.

To sum up the use case again:
Allow REST endpoints to be accessed without any call to authentication and with authentication present (Show different results and or allow/disallow writes to said endpoints based on hook.params.authenticated)

This is a requirement if you want to provide easy to use public service endpoints, it makes no sense to tell people to call empty authenticate before using it, its just to complicated.


daffl:
Shoot, sorry this has been sitting so long. Would you mind resolving the conflict and I can get it out with the next version?

MichaelErmer:
@daffl yeah lets do this, longest PR ever.

PR #600, opened 2017-11-03
Only set the JWT UUID if it is not already set

daffl:
This is a regression of #539. Basically if you were creating a new token using an existing JWT you always got an error since the `jti` property is already set in the original payload.

Closes https://github.com/feathersjs/authentication-jwt/issues/43


ekryski:
Ah so it is what I thought. Some people were recycling token payloads.

:shipit: Looks good. Thanks for fixing that @daffl!

xixilive:
👍

PR #604, opened 2017-11-08
Update handler.js

bertho-zero:
Fix this error:
```
        socket._feathers.body = {};
                              ^

TypeError: Cannot set property 'body' of undefined
```

### DIFF
--- a/lib/socket/handler.js
+++ b/lib/socket/handler.js
@@ -41,7 +41,7 @@ module.exports = function setupSocketHandler (app, options, { feathersParams, pr
         delete connection.accessToken;
         delete connection.authenticated;
         connection.headers = {};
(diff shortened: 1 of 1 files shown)


daffl:
When is that error happening?

bertho-zero:
- I initialize socket.io with autoConnect option set to false
- I open socket
- I run app.anthenticate() with expired token

bertho-zero:
Without this pull request I need to add the following code for my app to work:

```js
app.io.on('connection', socket => {
  if (!socket._feathers) {
    socket._feathers = {};
  }
}
```

daffl:
Ok, I'll try to reproduce and create a test case for it. I don't believe I ran into this so it's probably the `autoConnect` setting.

PR #610, opened 2017-11-21
Resolves #609 - Permit the developer to provide a custom authentication service implementation

### LINKED ISSUE #609: Allow custom authentication service registration
Pursuant to https://github.com/feathersjs/authentication/blob/master/lib/index.js#L52, permit the developer to provide their own authentication service.

thomas-p-wilson:
Resolves #609 by allowing the developer to specify a custom authentication service implementation as part of the options map passed to the authentication middleware.

I am not sure if there's any documentation I should update to accompany this. Please advise if this is necessary.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,9 +49,7 @@ function init (config = {}) {
       app.use(express.exposeCookies());
     }
 
-    // TODO (EK): Support passing your own service or force
-    // developer to register it themselves.
-    app.configure(service(options));
+    app.configure(options.authenticationService(options));
     app.passport.initialize();
 
     app.setup = function () {
--- a/lib/options.js
+++ b/lib/options.js
@@ -1,10 +1,12 @@
 const merge = require('lodash.merge');
+const service = require('./service');
 
 const defaults = {
   path: '/authentication',
   header: 'Authorization',
   entity: 'user',
   service: 'users',
+  authenticationService: service,
   passReqToCallback: true,
   session: false,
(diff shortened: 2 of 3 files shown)


Jaden-Giordano:
+1 on this, but I was thinking about the comment within the code, and wondering if it would be a good idea to just force the developer to register the service. Then we can update the generator to also generate the authentication service just like any other service. The programmer can just edit it or completely rewrite it. Just wanted some discussion to figure out which method we should implement because they both get the job done, but I think forcing the programmer to register the service themselves will decrease this codebase for this repo's size by about 50 lines or so and seems like in the long run would work better.

Navomi-Jaden:
@daffl What do you think?

daffl:
Thank you for the pull request @thomas-p-wilson. It is a nice addition, the question would be what a good use case would be (as opposed to e.g. setting `hook.result`). It's also risky because a custom auth service return value can mess with e.g. the client side authentication in potentially weird ways.

rottmann:
My use case would be a centralized account service, which will be used by all our apps. Like accounts.google.com. Each app need own accessTokens und within the token some stored data (e.g. roles, permissions, ...)

With the current solution we can only add account handling for one application.

refreshToken would also be nice.

daffl:
Closing in favour of Feathers v4 authentication. Please see the [migration guide](https://crow.docs.feathersjs.com/migrating.html) for more information.

PR #611, opened 2017-11-24
Options merged

Makingweb:
When trying to authenticate through sockets, options are not merged [(socket/handler.js#L82),](https://github.com/feathersjs/authentication/blob/master/lib/socket/handler.js#L82) such as hooks [hooks/authenticate.js#L55](https://github.com/feathersjs/authentication/blob/master/lib/hooks/authenticate.js#L55)

This can cause the following error if you try to define and use custom strategy:

`TypeError: Cannot read property 'entity' of undefined at service.create.then.tokens (@feathersjs\authentication\lib\socket\handler.js:95:31)`

### DIFF
--- a/lib/socket/handler.js
+++ b/lib/socket/handler.js
@@ -1,6 +1,8 @@
 const Debug = require('debug');
 const ms = require('ms');
 
+const merge = require('lodash.merge');
+
 const {
   normalizeError
 } = require('@feathersjs/socket-commons/lib/utils');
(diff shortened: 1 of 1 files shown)


daffl:
This makes sense, thank you and sorry for the delay! Will go out with the next version.

jamesholcomb:
I am curious if this fix would also address the intermittent "No auth token" `NotAuthenticated` errors I get over socket.io.  I use a custom passport strategy as well.

daffl:
I realized I forgot to make the release. Just published version 2.1.2 so you can give it a try.

PR #624, opened 2018-01-04
Allow the ability to specify a custom service to be used

Jaden-Giordano:
When passing the config to the configure function for authentication you can provide a function, similar to how most services are created, that registers a custom service under the property `service`.
```js
config.service = function(options) {
  return () => {
    app.use(options.path, {
      get: id => []
    });
  };
};

```
This will not register the original, default, service that is normally created, instead it will set the `authentication` service to the one specified under the `service` property within the config.

> Support passing your own service or force developer to register it themselves.

Reading that comment, which was originally within the code, I'm not completely sure if this the preferred solution, but it is _a solution_; if that's the case then I'm willing to make any changes to move to a more preferred method.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,9 +49,11 @@ function init (config = {}) {
       app.use(express.exposeCookies());
     }
 
-    // TODO (EK): Support passing your own service or force
-    // developer to register it themselves.
-    app.configure(service(options));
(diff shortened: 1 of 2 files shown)


Jaden-Giordano:
I'm slow... Closing this because its already been implemented #610.

PR #652, opened 2018-03-16
Fix setTimeout memory leak

wnxhaja:
### Steps to reproduce
- login user -> logout user
- login user -> user disconnected ->  re-authenticate user

### Expected behavior
- previous logoutTimer (setTimeout) should be cleared

### Actual behavior
- logoutTimer still stays

### Reason
- this one will never be true https://github.com/feathersjs/authentication/blob/master/lib/socket/handler.js#L110 if the user was disconnected then tried to re-authenticate.

-  should clear the timeout here too https://github.com/feathersjs/authentication/blob/master/lib/socket/handler.js#L59 when socket is disconnected and user tried to reauthenticate.

screen shot 2018-03-16 at 5 05 26 pm

screen shot 2018-03-16 at 5 06 17 pm

### DIFF
--- a/src/socket/handler.js
+++ b/src/socket/handler.js
@@ -23,10 +23,9 @@ export default function setupSocketHandler (app, options, { feathersParams, prov
   const service = app.service(authSettings.path);
   const entityService = app.service(authSettings.service);
   let isUpdateEntitySetup = false;
+  let logoutTimer;
 
   return function (socket) {
-    let logoutTimer;
-
     const logout = function (callback = () => {}) {
       const connection = feathersParams(socket);
       const { accessToken } = connection;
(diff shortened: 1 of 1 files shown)


daffl:
Could you re-submit this against master?

wnxhaja:
I did here https://github.com/feathersjs/authentication/pull/653

daffl:
Cool. This one is closed via #653 then.

wnxhaja:
@daffl can you fix the `auk` too because we have projects running on that version.

daffl:
Other than Node 4 support, `@feathersjs/authentication` and related packages did not change so they should be backwards compatible and can be upgraded individually.

Eventually you'd probably want to upgrade anyway. The process is mostly automated and should be fairly quick. So far we haven't heard of anybody having major issues with it and it'll reliably get you the latest versions of everything.

PR #664, opened 2018-04-12
Return the original accessToken for jwt authentication

daffl:
It is possible to authenticate using JWT which currently returned a new token. The original intention was to validate and return the existing token.

For https://github.com/feathersjs/authentication-jwt/issues/61


daffl:
Closing this since it will be addressed more thoroughly in the next version.

PR #665, opened 2018-04-15
Remove payload and user entity on logout.

bertho-zero:
Without this change, the services recognize users even after logging out, until the page is reloaded or the socket is forced to reconnect.


daffl:
Thank you!

PR #685, opened 2018-06-28
Replaced feathers.static with express.static

georgehorrell:
### Summary

The quick example contained in the README fails to run. In Buzzard, `feathersjs/express` contains the Express built-in middleware like `express.static`. This change updates the README to match the new Buzzard API.

There are no open issues related to this change, and this change is not dependent on any other PRs.


daffl:
Indeed, thank you!

PR #686, opened 2018-06-28
Remove subject from the JWT verification options

rasendubi:
As discussed in #684, the current default for subject causes token
verification to fail if sub is overriden.

Removing subject from default options might be a breaking change as
people might rely on it.

Remove subject from verification options, so it's not checked. This is
a safer way.


daffl:
Great, thank you! Released as v2.1.7

rasendubi:
Thank you!

